PR #10, opened 2015-07-03
ensure docker image complies with eris platform stds

compleatang:
This PR performs requests the following changes be made:

* update the docker image to comply with the current eris platform paradigms.
* as discussed on Slack, we want the erisdb:0.10 container to be compiled with the following binaries:
    - tendermint
    - erisdb
    - barak
    - debora
    - erisdbss
* this will make for a heavy image, but as most of the image's "weight" is in the OS layer it is really only a difference of about 30MB (in the grand scheme of things that's not a lot). the advantage of compiling all of these binaries is that because all of the data can be mint based we are able to put the image in `cli` mode and turn on its `rpc` via the proper mint binary, or in `server` mode and turn on the `rest|stream` api via erisdb, or in `test` mode and turn on the `erisdbss`. this will make for a powerful and flexible image.
* the last change, which is making the tests fail and so will need some work before merging into develop is an update to mint to the current head of tendermint permissions branch. 
* updates circle to build and push an image when we push to master.

**Note** the start script may need some massaging to make this work seamlessly with the platform.


DOCKER/start.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+if [[ $BARAK_SEED ]]; then
+  cat /go/src/github.com/eris-ltd/eris-db/Godeps/_workspace/src/github.com/tendermint/tendermint/cmd/barak/$BARAK_SEED | barak &
```
ebuchman: this doesn't work since tendermint/tendermint/cmd is not a dependency (its not vendored - godeps is at least some what intelligent)

ebuchman: but anyways, we shouldnt run barak in the container i don't think...

compleatang: duh. ye, mint is there in the container, but i'm reworking this now and adding mint-client so will update....

DOCKER/Dockerfile
```diff
@@ -15,9 +16,34 @@ RUN apt-get update && \
...
+  cd ../erisdbss && \
+  go build && \
+  mv erisdbss /usr/local/bin/
+
+# persist data, set user
+VOLUME /home/$USER/.eris
+WORKDIR /home/$USER/.eris
+USER $USER
+RUN mkdir --parents /home/$USER/.eris/blockchains/tendermint
+ENV TMROOT /home/$USER/.eris/blockchains/tendermint
```
ebuchman: this would be fine if we only have one blockchain in a container at a time, but what if we want to share a data container with the entire ~/.eris across service and chain containers?  then this will have to be something like blockchains/chainName/chainNumber, and we cant set it in the Dockerfile

compleatang: sure. but it's just a default. since the whole `.eris` is in a volume we can use data containers there and then just set the TMROOT env var in the chain definition file.... at least that was my thinking....

PR #48, opened 2016-01-16
WIP tmint

benjaminbollen:
move eris/tendermint into eris-db; replaying work done on eris-db tmsp branch for current develop branch

### DIFF
--- a/account/signature_test.go
+++ b/account/signature_test.go
@@ -1,0 +1,70 @@
+package account
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/tendermint/go-ed25519"
+	"github.com/tendermint/go-wire"
+	. "github.com/tendermint/go-common"
+)
+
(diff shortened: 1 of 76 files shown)


account/signature_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+	buf, n, err := new(bytes.Buffer), new(int), new(error)
+	wire.WriteBinary(sig, buf, n, err)
+	if *err != nil {
+		t.Fatalf("Failed to write Signature: %v", err)
+	}
+
+	if len(buf.Bytes()) != ed25519.SignatureSize+1 {
+		// 1 byte TypeByte, 64 bytes signature bytes
+		t.Fatalf("Unexpected signature write size: %v", len(buf.Bytes()))
```
benjaminbollen: FAILS: buf is 64 bytes long; seems type byte is missing.  Will follow up

PR #68, opened 2016-03-11
Fixes

ebuchman:
- lots of little fixes in network connectivity (don't filter by ip range, make all addresses routable, only connect to peers with same genesis hash)
- upgraded mempool to latest tendermint version
- some consensus bug fixes


benjaminbollen:
looks (diagonally) good to me

PR #92, opened 2016-03-30
WIP Upgrade tendermint to a first class citizen in eris-db

benjaminbollen:
the tendermint tests fail; were not being run before.


circle.yml
```diff
@@ -17,7 +17,7 @@ dependencies:
     - sudo service docker start
     - "sudo apt-get update && sudo apt-get install -y libgmp3-dev"
     - sudo curl -sSL -o /usr/bin/docker-machine https://github.com/docker/machine/releases/download/v${DOCKER_MACHINE_VERSION}/docker-machine_linux-amd64; sudo chmod 0755 /usr/bin/docker-machine
-    - docker login -e $DOCKER_EMAIL -u $DOCKER_USER -p $DOCKER_PASS quay.io
+    # - docker login -e $DOCKER_EMAIL -u $DOCKER_USER -p $DOCKER_PASS quay.io
```
benjaminbollen: This is a poor solution to make sure CI does not hang on this.  Needs to be removed from the git history before moving down to develop.

silasdavis: Shall I try and do a separate CircleCI PR to fix this?

benjaminbollen: yes, possibly with a small script we can circumvent this; at least that's the best idea I have

PR #108, opened 2016-05-06
move files under txs/ into txs package

silasdavis:
Minor clean-up exercise, but we would like to follow package naming conventions, particularly since it breaks some tooling to not.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -39,6 +39,7 @@ test:
   pre:
     - go get github.com/stretchr/testify
     - go get github.com/Masterminds/glide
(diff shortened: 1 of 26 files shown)


benjaminbollen:
Lets have full word naming conventions

silasdavis:
@benjaminbollen what are you suggesting, to rename txs package to transactions?

benjaminbollen:
yes, but we don't have to do this now; I mentioned it as a coding style Id prefer moving forward

benjaminbollen:
fails on tests, but looks good to me

PR #116, opened 2016-05-11
Start of CHANGELOG.md

benjaminbollen:
Propose a CHANGELOG.md; format borrowed from go-ipfs

changelog should be updated on every release cycle; as a backlog item we can complete for recent versions.

apologies for also pulling in minor clean-up, if any objections happy to separate out the commits.

### DIFF
--- a/erisdb/params.go
+++ b/erisdb/params.go
@@ -90,13 +90,6 @@ type (
 		GasLimit int64  `json:"gas_limit"`
 	}
 
-	// Used when sending a 'Send' transaction.
-	SendParam struct {
-		PrivKey   []byte `json:"priv_key"`
-		ToAddress []byte `json:"to_address"`
-		Amount    int64  `json:"amount"`
-	}
-
 	NameRegEntryParam struct {
 		Name string `json:"name"`
 	}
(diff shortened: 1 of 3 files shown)


erisdb/params.go
```diff
@@ -90,13 +90,6 @@ type (
 		GasLimit int64  `json:"gas_limit"`
 	}
 
-	// Used when sending a 'Send' transaction.
-	SendParam struct {
```
silasdavis: Why's this going?

silasdavis:
Otherwise LGTM and will merge

erisdb/params.go
benjaminbollen: because in TMSP refactor `Send` and `SendAndHold` had already been removed, and in the merge this struct redundantly lingered (noticed when I manually ran over the diff).  Arguably this will be reverted if `Send` and `SendAndHold` are not to be deprecated, but to not add to the amount of loose ends, I choose to for now also delete it.

In https://github.com/eris-ltd/eris-db/issues/115 we can discuss what the fate of `Send` and `SendAndHold` is.

PR #121, opened 2016-05-16
[DO NOT MERGE] Tmroot

benjaminbollen:
- [ ] 1. bring e-cm and e-db in line
- [x] 2. move tendermint to the status of a consensus module
- [x] 3. clean up config and command line interface for eris-db with cobra and viper
- [ ] 4. clean up DOCKER and shell scripts interface into eris-db

out-of-scope extras
a. import event subscription infrastructure from go-ethereum


DOCKER/start.sh
```diff
@@ -41,7 +41,7 @@ fi
 CHAIN_DIR="/home/$USER/.eris/chains/$CHAIN_ID"
 
 # set the tendermint directory
-TMROOT=$CHAIN_DIR
+TMROOT="$CHAIN_DIR"/tendermint
```
benjaminbollen: typo noted

event/event.go
```diff
@@ -0,0 +1,241 @@
...
+import (
+	"errors"
+	"fmt"
+	"reflect"
+	"sync"
+	"time"
+)
+
+// Event is a time-tagged notification pushed to subscribers.
+type Event struct {
```
silasdavis: So we're copying this in, not depending on Ethereum for it? I can see a rationale as it changes for different reasons and it's a fairly clean piece of code, but worth considering the alternative.

benjaminbollen: True, worthwhile consideration but i am of the opinion that here we are better served taking ownership of this code:
1. As you mentioned it is a clean, no extra dependencies, clean piece of code - thank you geth
2. The event switch is absolutely crucial for an event based architecture; so i see many dangers considering the event switch third party code

silasdavis: OK, I'm sold.

definitions/do.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func NowDo() *Do {
+	do := new(Do)
+	do.Debug = false
+	do.Verbose = false
+	// the default value for output is set to true in cmd/eris-db.go;
+	// avoid double setting it here though
+	do.Output = false
+  do.ChainId = ""
```
silasdavis: indent

definitions/do.go
```diff
@@ -0,0 +1,52 @@
...
+  viper "github.com/spf13/viper"
+)
+
+type Do struct {
+  // Persistent flags not reflected in the configuration files
+  // only set through command line flags or environment variables
+	Debug        bool     // ERIS_DB_DEBUG
+	Verbose      bool     // ERIS_DB_VERBOSE
+	Output       bool     // ERIS_DB_OUTPUT
+  // Capital configuration options explicitly extracted from the Viper config
```
silasdavis: You mean they appear here an in config? How do we protect from divergence?

cmd/eris-db.go
```diff
@@ -0,0 +1,133 @@
...
+)
+
+const VERSION = version.VERSION
+
+// Global Do struct
+var do *definitions.Do
+
+var ErisDbCmd = &cobra.Command {
+  Use:   "eris-db",
+  Short: "Eris-DB is the beating heart of the eris chain.",
```
silasdavis: Emosh

definitions/do.go
benjaminbollen: no debug, verbose and output are not set in config; flags that are reflected in config will have to be covered by viper

cmd/eris-db.go
benjaminbollen: made with <3

silasdavis:
so this LGTM, but assuming WIP means you don't want it merged yet

benjaminbollen:
Closing this PR as it is not ready to move into `develop-tmsp`.  These commits are pushed at branch `discuss-eris-gateway` as the essentials of ErisMint as an application are tied up with Tendermint as an in-process consensus engine.
At this point the E-DB servers need to be loaded in this new structure and the RPC calls need to be separated to application and consensus behind the pipe.

PR #122, opened 2016-05-16
Improve glide dependencies using fire

silasdavis:
This PR removes the transitive dependencies of tendermint/tendermint, and just depends on tendermint/tendermint, allowing glide to do its thing and pull in the various tendermint dependencies as specified by tendermint's glide.yaml. This should give us a consistent set of versions of tendermint dependencies.


benjaminbollen:
where is the :fire: ?

silasdavis:
Raw emoji meant to be there... okay without :fire: this time?

silasdavis:
I've removed the subpackages directive, but I will reintroduce if there is a reason to do so, but I think we can merge now anyway (see: https://github.com/Masterminds/glide/issues/424)

glide.yaml
```diff
@@ -1,32 +1,10 @@
...
-- package: github.com/tendermint/go-config
-- package: github.com/tendermint/go-crypto
-- package: github.com/tendermint/go-db
-- package: github.com/tendermint/go-events
-- package: github.com/tendermint/go-logger
-- package: github.com/tendermint/go-merkle
-- package: github.com/tendermint/go-p2p
 - package: github.com/stretchr/testify
-- package: github.com/tendermint/go-rpc
-- package: github.com/tendermint/go-wire
```
benjaminbollen: will we not have explicit dependencies on go-wire still, and others of these

glide.lock
```diff
@@ -1,8 +1,8 @@
-hash: 9cba3c71b5ea9762cc27faa683ad5611c47de7761f5633131b92cde7961eb227
-updated: 2016-05-11T16:58:37.071884134-04:00
+hash: 21311609a866f185014cc1fead70d1a2898f0ccbd45244b81401ec72eaab39a0
+updated: 2016-05-16T20:08:05.365799111+01:00
 imports:
 - name: github.com/btcsuite/btcd
```
benjaminbollen: where does this dependency come from?

glide.yaml
silasdavis: yes, but I think we want to hook ourselves to tendermint's version because we'll be relying on whatever they are doing 'on their side'

benjaminbollen: the use of go-wire in the state of edb is logically disconnected from the use of go-wire inside of tendermint, but this should not be a pressing issue today

glide.yaml
```diff
@@ -1,6 +1,7 @@
 package: github.com/eris-ltd/eris-db
 import:
 - package: github.com/tendermint/tendermint
+  version: 55ef4b225fb0aa5637a96d36a8c0d030a59dc21d
```
benjaminbollen: it will be interesting to see how this plays out with the independent tendermint/go-* repositories, but :+1:

benjaminbollen:
looks good to me

PR #124, opened 2016-05-17
Test build fixes

ebuchman:
some simple dependency fixes to make tests build.

note the `test` folder is out of scope - we're moving to deprecate it because I don't really know how anyone is expected to be able to maintain it (very brittle). and of course rpc tests are still failing, ala at least https://github.com/eris-ltd/eris-db/issues/109


ebuchman:
oops

PR #126, opened 2016-05-17
Tests building considered success

silasdavis:
So this is a kind of filthy one guys... But for the time being, so that Circle CI can actually provide us with some useful feedback on PRs (other than the tests are failing - thanks, we know) I'm proposing we just build the tests, and don't run them. We can add in those tests that we fix to run as we fix the piecemeal.

So this:

- Nukes ./test/... (good idea?) in favour of starting from scratch on those against refactor
- Overrides circle's test to just build the tests, while we work towards working tests


silasdavis:
@ebuchman tagging you for heads up

silasdavis:
@benjaminbollen let me know what you think

benjaminbollen:
1. we don't have to delete anything, we can rename to `_tes.go` instead of `_test.go`
2. I would want the tests that currently pass to still be run

rpc/core/1
```diff
@@ -1,55 +0,0 @@
-package core
```
benjaminbollen: yes, why this was committed in the first place I wondered about today as well :)

ebuchman:
what non passing tests are left, besides rpc? 
lets take all failing nonrpc to _tes.go, and just be explicit for now in the circle which package tests to run and which specific tests of rpc to run. and as we fix up those rpc tests we can add them until the whole package is fixed

silasdavis:
So we don't need to rename any tests, I am just compiling the tests that are there.

I deleted test/ independently because it doesn't compile and rather than fix that I know we have questioned its value. I can reinstate test/ if that was too rash.

benjaminbollen:
while /test isnt ideal; I do favor:
1. not deleting it; and it should compile
2. I prefer disabling tests over deleting them as they serve as a placeholder for what needs to be there

silasdavis:
Yeah I am not suggesting deleting tests in general. I'll reinstate test/

benjaminbollen:
will close this

PR #136, opened 2016-06-10
finalize the right padding of permissions for other roles funcs

compleatang:
don't merge just yet. testing locally.


compleatang:
OK. Tests now passing. GTM. Will open another against develop-tmsp which captures the changes. Don't think we have hotfixed master, but I've pushed the changes to quay manually and we're about to move to rc1 at any rate so I think it'll be a waste of time to PR to master.

benjaminbollen:
LGTM

PR #137, opened 2016-06-10
Native support for ARM

shuangjj:
closes https://github.com/eris-ltd/eris-cli/issues/718


zramsay:
LGTM, @benjaminbollen @silasdavis can we get a :+1: for the changes in the edb binary

erisdb/pipe/blockchain.go
```diff
@@ -1,3 +1,5 @@
+// +build !arm
```
benjaminbollen: what does this line do?

erisdb/pipe/blockchain_arm.go
```diff
@@ -0,0 +1,269 @@
+package pipe
```
benjaminbollen: what is different in blockchain_arm.go?

benjaminbollen:
As an additional point, e-db is undergoing a restructuring, so we will have to port these changes to the new branches (ideally discuss-eris-gateway), but we can take this onboard as we make the transition

benjaminbollen:
The changes in blockchain.go refer to `MaxUInt32` which is replaced by `MaxInt32`?  This only occurs in this file?

erisdb/pipe/blockchain_arm.go
```diff
@@ -0,0 +1,269 @@
...
+	match func(int, int) bool
+}
+
+func (this *BlockHeightFilter) Configure(fd *FilterData) error {
+	op := fd.Op
+	var val int
+	if fd.Value == "min" {
+		val = 0
+	} else if fd.Value == "max" {
+		val = math.MaxInt32
```
shuangjj: `var val int` should be signed 32-bit integer, but the assignment `val = math.MaxUint32` will overflow it. I tried to change the val type to uint32, but that would incur many other compilation errors. So the workaround for 32-bit ARM is to change the assignment to `val = math.MaxInt32`.

benjaminbollen:
Please note that this will be overwritten as these files have moved in the restructuring of the consensus and pipe code.  The preferred solution will be to use `uint64` (or `uint32` if someone wants to argue for that as a max height), and replace the types consistently everywhere.

@zramsay thanks for linking this PR in a new issue

PR #138, opened 2016-06-10
fixup for develop-tmsp

compleatang:


benjaminbollen:
LGTM; copies #136 into develop-tmsp

benjaminbollen:
I also added these changes to my local branch of `discuss-eris-gateway`, so they should not get lost

PR #140, opened 2016-06-10
add git verify-commit HEAD on CI

benjaminbollen:


benjaminbollen:
if people like this proposal, please comment here on which key servers need to be added to the list:
- pgp.mit.edu

benjaminbollen:
@silasdavis we don't have to merge this but we can use it as a discussion starter on what we would want to adhere to as a policy towards signing commit history

silasdavis:
This seems like a good idea, but what are we actually trying to achieve?

Seems like we need a human-maintained whitelist of keys, where we can check identities against something like keybase.io and have some degree of belief that commits link to a somewhat known identity.

This might fit in well if we decide that we need a contributor license agreement (CLA) to be signed by individuals contributing.

Another general advantage is that we only add a specific key, and we don't end up with commits from unknown keys being mixed in with a pull request seemingly from a known identity. I wouldn't imagine us trying to do a very deep audit on identities as this would be onerous and somewhat against the spirit of open source.

benjaminbollen:
Yes, I think the biggest danger here is putting up a barrier for other people to start contributing code, and that is what we absolutely want, not the opposite.

So maybe we can have more stringent checks on merging in commits into develop or master (but then those need to happen locally, and we would lose the GH gui, a minor inconvenience)

AFDudley:
Yeah, it seems like anyone can submit a pull request, unsigned. But whomever audits it on the Eris side will have to sign their merge? I don't know how that would work within the git flow model, but that seems like the best of both worlds.

benjaminbollen:
This PR does not allow that, so we can change it to assert the signatures of all merge commits

benjaminbollen:
I needed my `develop` branch so I have parked these commits on https://github.com/benjaminbollen/eris-db/tree/check_signatures_henceforth

PR #142, opened 2016-06-14
Activate servers

benjaminbollen:
We move the servers as they exist from Eris-DB into a new structure; but we explicitly preserve the existing code and API.  The aim is to subsequently review collectively a new architecture for the API.

Note that this PR does not yet have the Tendermint exposed RPC

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -58,7 +58,9 @@ func NewCore(chainId string, consensusConfig *config.ModuleConfig,
 	}
 	log.Debug("Loaded pipe with application manager")
 	// pass the consensus engine into the pipe
-	consensus.LoadConsensusEngineInPipe(consensusConfig, pipe)
+	if e := consensus.LoadConsensusEngineInPipe(consensusConfig, pipe); e != nil {
+		return nil, fmt.Errorf("Failed to load consensus engine in pipe: %v", e)
+	}
 	tendermintPipe, err := pipe.GetTendermintPipe()
 	if err != nil {
 		log.Warn(fmt.Sprintf("Tendermint gateway not supported by %s",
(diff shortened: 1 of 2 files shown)


rpc/core/pipe.go
```diff
@@ -1,61 +1,60 @@
-package core
-
-import (
-	stypes "github.com/eris-ltd/eris-db/manager/eris-mint/state/types"
-	"github.com/eris-ltd/eris-db/tmsp"
```
benjaminbollen: This line caused glide to self-vendor eris-db

benjaminbollen:
I needed my `develop` branch so I have parked these commits on https://github.com/benjaminbollen/eris-db/tree/check_signatures_henceforth

benjaminbollen:
This PR now does have the outlines for the Tendermint exposed routes; implementations now have to move to the application managers implementation of the RPC calls (for /rpc/tendermint these are still defined in /rpc/tendermint/core)

core/core.go
```diff
@@ -58,7 +58,10 @@ func NewCore(chainId string, consensusConfig *config.ModuleConfig,
 	}
 	log.Debug("Loaded pipe with application manager")
 	// pass the consensus engine into the pipe
-	consensus.LoadConsensusEngineInPipe(consensusConfig, pipe)
+	err = consensus.LoadConsensusEngineInPipe(consensusConfig, pipe)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to load consensus engine in pipe: %v", err)
+	}
```
zramsay:
```
if err := consense.L(); err !=nil {
        return nil, fmt.Errorf()
}
```

benjaminbollen: good point here

PR #150, opened 2016-07-05
eris gateway

benjaminbollen:
Eris-DB has to evolve from an explicit chain implementation to a gateway functionality such that an increasingly rich landscape of blockchain technologies can be normalised through a modular gateway.

- introduce separation of consensus engine, application manager, and gateway rpc
- support tendermint 0.6 in-process
- support ErisMint 0.12 in-process
- support Eris-DB RPC; Tendermint RPC


silasdavis:
So... I'm not sure I'm going to code review this as is... this is organisational right?

PR #157, opened 2016-07-11
Issue152 disable mock web api tests

benjaminbollen:
Disable mock web_api tests; blocked on Codec / go-wire Parsing issue; but also tests need a separation of actual rpc test and state test


silasdavis:
Roger that, MergeyMcMergeFace

PR #160, opened 2016-07-11
Deployment docker and ecm

benjaminbollen:
includes #157 partially, so merge after #157 is merged.


DOCKER/Dockerfile
```diff
@@ -1,80 +0,0 @@
...
-
-# install glide for dependency management
-RUN	go install github.com/Masterminds/glide
-
-# set the repo and install erisdb
-ENV REPO $GOPATH/src/github.com/eris-ltd/eris-db
-WORKDIR $REPO
-COPY . $REPO
-RUN glide install
-RUN cd ./cmd/eris-db && go build -o /usr/local/bin/eris-db && cd ../..
```
silasdavis: I don't think the final cd ../.. is needed; RUN always executes to whatever WORKDIR is set to

silasdavis: I just commented on a deleted line.

tests/build_tool.sh
```diff
@@ -0,0 +1,50 @@
...
+then
+  repo=`pwd`
+else
+  base=github.com/eris-ltd/eris-db
+  repo=$GOPATH/src/$base
+fi
+branch=${CIRCLE_BRANCH:=master}
+branch=${branch/-/_}
+
+release_min=$(cat $repo/version/version.go | tail -n 1 | cut -d \  -f 4 | tr -d '"')
```
silasdavis: Minor but would this be less fragile if we added a $repo/version/main.go file that prints the version then we could do `$(go run $repo/version)` to get the version

silasdavis:
LGTM

tests/build_tool.sh
benjaminbollen: good comment, but this should be done across the stack for consistency reasons and there is an argument for using solid bash. definitely worth raising an issue and following up

PR #170, opened 2016-07-15
Add receipts and fix BroadcastTx tests

### LINKED ISSUE #109: return a tx receipt from CheckTx/AppendTx
Return https://github.com/eris-ltd/eris-db/blob/develop-tmsp/rpc/core/types/responses.go#L102 and 
https://github.com/eris-ltd/eris-db/blob/develop-tmsp/tmsp/erisdb.go#L145
ought to return https://github.com/eris-ltd/eris-db/blob/develop-tmsp/rpc/core/types/responses.go#L102
so that https://github.com/eris-ltd/eris-db/blob/develop-tmsp/rpc/test/tests.go#L81 works.

Note the receipt used to be generated in the broadcast_tx rpc function itself (https://github.com/eris-ltd/tendermint/blob/develop/rpc/core/mempool.go#L29) but broadcast_tx is now going to be something that happens on tendermint's end, so won't necessarily have access to that unless it's returned as data or log in the TMSP response. alternatively, if we keep broadcast_tx in eris_db, it could happen in broadcast_tx, but the erisdb tmsp app would be more useful if it returned this data itself

silasdavis:
With a bit of related refactoring.

May substantive changes are `GenerateReceipt` and callers thereof. The receipt gets marshalled and unmarshalled from the data payload of the Tendermint response.

Closes #109

### DIFF
--- a/account/account.go
+++ b/account/account.go
@@ -28,7 +28,6 @@ import (
 	ptypes "github.com/eris-ltd/eris-db/permission/types"
 	. "github.com/tendermint/go-common"
 	"github.com/tendermint/go-crypto"
-	"github.com/tendermint/go-merkle"
 	"github.com/tendermint/go-wire"
 )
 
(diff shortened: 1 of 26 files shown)


benjaminbollen:
`FAIL	github.com/eris-ltd/eris-db/txs [build failed]`

silasdavis:
Oops, yep fixed

benjaminbollen:
LGTM

PR #173, opened 2016-07-19
Fix Dockerfile; install dependencies with glide missing

benjaminbollen:
I forgot to actually install the dependencies with glide ...


benjaminbollen:
will rebase

PR #177, opened 2016-07-21
Implement Subscription and other missing things

silasdavis:
Makes some tests pass in rpc/tendermint/test, not fully, but better than before. Small pieces :cake:


benjaminbollen:
LGTM

Issue #164, opened 2016-07-14
Account address generation for factory pattern can lead to storage overwrite of existing accounts 


j-h-scheufen:
```
OSX version: 10.11.5
Eris CLI Version: 0.12.0
Docker:
Client:
 Version:      1.11.1
 API version:  1.23
 Go version:   go1.5.4
 Git commit:   5604cbe
 Built:        Tue Apr 26 23:44:17 2016
 OS/Arch:      darwin/amd64
Server:
 Version:      1.11.1
 API version:  1.23
 Go version:   go1.5.4
 Git commit:   5604cbe
 Built:        Wed Apr 27 00:34:20 2016
 OS/Arch:      linux/amd64
docker-machine version 0.7.0, build a650a40
```
The following problematic behavior has been observed when creating new contract instances from within other contracts in Solidity:
1. Calling the `new` constructor of another contract from a *top-level* contract account, i.e. the contract that is invoked by an external call from EPM or JS, always creates a new address as expected.
2. Delegating from the *top-level* contract to another contract (or further down) to call the `new` constructor creates a new address only upon first invocation. Any subsequent invocation results in creating the same address and as a result the storage that existed at the address is overwritten!

The attached contracts and EPM scripts ([factory-test.zip](https://github.com/eris-ltd/eris-db/files/364748/factory-test.zip)) can be used to reproduce the behavior in the following sequence:

1. Create a new chain
2. `cp epm.yaml.1 epm.yaml` and run it **once** to deploy the contracts
3. Copy the address of the deployed `InstructionManager` contract (which acts as the mentioned *top-level contract*) from the console output and paste it in the first job in the `epm.yaml.2` file
4. `cp epm.yaml.2 epm.yaml` and run it **once**.
5. Take note of the new Instruction contract addresses created by the *level 1 & 2* factory contracts. Use these addresses to make a copy of their storage, e.g.: `http://192.168.99.100:46657/dump_storage?address="799DF4C5232A8CB99F8A5237B999B8C84DBF7746"`
6. Run the epm.yaml **again**. This time around it will provoke the creation of already existing addresses by the *level 1 & 2* factories and the storage at these addresses will be overwritten.

The following is the console output from the above sequence:
```
$ cp epm.yaml.1 epm.yaml

jan at MacBook-Pro-6 in ~/Development/git/github.com/eris-ltd/eris-contracts-bundles/src/factory-test on instruction-registry-indexing-48 [!?]
[...]

PR #181, opened 2016-07-28
tendermint vm: CREATE opcode pass NewAccount caller

benjaminbollen:
relates to (and solves on master) #164;
fix indicated by reduced tests posted by @dennismckinnon to be integrated in eris-pm

PRs follow for eris-db:develop


benjaminbollen:
callee in vm is the current contract being executed; while this makes the test pass, it is not the correct answer to the problem.

PR #182, opened 2016-07-28
Create fix

ebuchman:
So this was actually pretty bad. Amazing that we never got it, probably because we never really cared about tokens. Basic "address.send()" from a contract [was broken](https://github.com/eris-ltd/eris-db/compare/create_fix?expand=1#diff-735ecc7ace7329b12bfb59eaa38f682bR495)! [This comment/contract](https://github.com/eris-ltd/eris-db/blob/develop/manager/eris-mint/evm/vm.go#L116) wasn't true, since we were [making calls without adding the account to the tx cache](https://github.com/eris-ltd/eris-db/blob/develop/manager/eris-mint/evm/vm.go#L814).

The new tests involve pre-compiled solidity contracts. It may be better to move them to epm, or to replace them with hand-written minimal byte code that captures the functionality.

### DIFF
--- a/manager/eris-mint/evm/vm.go
+++ b/manager/eris-mint/evm/vm.go
@@ -735,6 +735,7 @@ func (vm *VM) call(caller, callee *Account, code, input []byte, value int64, gas
 			// TODO charge for gas to create account _ the code length * GasCreateByte
 
 			newAccount := vm.appState.CreateAccount(callee)
+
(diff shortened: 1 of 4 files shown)


benjaminbollen:
Looks good to me;

At cost of greater overhead we can update at the call itself, but should be updated now as persisted state is nonce, balance, code and storage.

PR #187, opened 2016-08-01
eris-client: carry-over from mint-client; client/core lib for eris-pm

benjaminbollen:
- [x] eris-client basic framework carry-over
- [x] eris-client tx send
- [x] call/bond/...


VoR0220:
one thing this needs is a getChainID method.

VoR0220:
or a chain status method that only needs the address of the chain.

benjaminbollen:
additional requirements (other than given by mint-client) will be tracked in #264

benjaminbollen:
has transaction factory in `client/core` for Send, Call, Name, and Permissions; with very simple test framework for construction.  This unit test framework can significantly be improved, and should be.

Tomorrow I will open a new PR to expose the RPC functions that are needed for e-pm @VoR0220 ; see line comment for `GetAccount`

client/core/transaction_factory.go
```diff
@@ -0,0 +1,450 @@
...
+// wait for events
+
+type Msg struct {
+	BlockHash []byte
+	Value     []byte
+	Exception string
+	Error     error
+}
+
+// func subscribeAndWait(tx txs.Tx, chainID, nodeAddr string, inputAddr []byte) (chan Msg, error) {
```
silasdavis: Can we add a TODO here, perhaps reference similar functionality in: 

https://github.com/eris-ltd/eris-db/blob/develop/rpc/tendermint/test/ws_helpers.go#L137

Might be a good idea to share implementation... might not...

keys/key_client_util.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+// Eris-Keys server connects over http request-response structures
+
+type HTTPResponse struct {
+	Response string
+	Error    string
+}
+
+func RequestResponse(addr, method string, args map[string]string) (string, error) {
```
silasdavis: This is a bit non-standard. Rename to 'MakeRequest', or even 'Do'. Unless I'm missing something this is just a wrapping of making a HTTP request. I don't know of any libraries that would call this 'RequestResponse', despite there being a certain mind-bending logic to it...

client/client.go
```diff
@@ -0,0 +1,82 @@
...
+	"fmt"
+
+	"github.com/tendermint/go-rpc/client"
+
+	"github.com/eris-ltd/eris-db/account"
+	tendermint_client "github.com/eris-ltd/eris-db/rpc/tendermint/client"
+	"github.com/eris-ltd/eris-db/txs"
+)
+
+type NodeClient interface {
```
benjaminbollen: I agree names are important, but I do like `Node` for the reason I call it here `ErisNodeClient` and `ErisKeyClient`; and yes `TendermintNode` (it is a node on its own at that consensus level);

`Node` in `ErisNodeClient` is a running instance of eris-db

client/core/transaction_factory.go
```diff
@@ -0,0 +1,450 @@
...
+	// args := snativeArgs(
+	tx := txs.NewPermissionsTxWithNonce(pub, args, int(nonce))
+	return tx, nil
+}
+
+// type NameGetter struct {
+// 	client cclient.Client
+// }
+
+// func (n NameGetter) GetNameRegEntry(name string) *txs.NameRegEntry {
```
benjaminbollen: added todo in https://github.com/eris-ltd/eris-db/issues/264

client/core/transaction_factory.go
benjaminbollen: this is directly to be continued; Im still on this

keys/key_client_util.go
benjaminbollen: these utils are an exact copy of mintx, see https://github.com/eris-ltd/mint-client/blob/master/mintx/core/core.go#L320-L372

Noted arguments and this can be reduced to a reusable component; nothing here is earth shattering

client/core/transaction_factory.go
benjaminbollen: thanks for pointer as well, because the exact reason I left it commented was that I needed to find where I could re-use the code

silasdavis:
LGTM

PR #222, opened 2016-08-16
Combine consensus and manager events into a pool of types we control

silasdavis:
The approach here is to deal with underlying go-events EventData that is a generic marker interface by mapping it to our our own EventData types, which is a marker interface with an assertion method. This is in the spirit of the refactor to take control of the types we depend on. Ultimately we still care about some tendermint types but we are reducing the surface area of the dependency.

This resolves some issues with subscription, which exposes some other issues that we will fix with a subsequent PR.

### DIFF
--- a/event/events.go
+++ b/event/events.go
@@ -21,7 +21,11 @@ import (
 	"encoding/hex"
 	"strings"
 
-	evts "github.com/tendermint/go-events"
+	"fmt"
+
+	"github.com/eris-ltd/eris-db/txs"
+	go_events "github.com/tendermint/go-events"
+	tm_types "github.com/tendermint/tendermint/types"
 )
 
 // TODO improve
@@ -29,12 +33,15 @@ import (
 // [ben] To improve this we will switch out go-events with eris-db/event so
 // that there is no need anymore for this poor wrapper.
 
+// Oh for a sum type
+type anyEventData interface{}
+
 type EventEmitter interface {
-	Subscribe(subId, event string, callback func(evts.EventData)) error
+	Subscribe(subId, event string, callback func(txs.EventData)) error
 	Unsubscribe(subId string) error
 }
 
-func NewEvents(eventSwitch *evts.EventSwitch) *events {
+func NewEvents(eventSwitch *go_events.EventSwitch) *events {
 	return &events{eventSwitch}
 }
 
@@ -47,12 +54,21 @@ func Multiplex(events ...EventEmitter) *multiplexedEvents {
 
 // The events struct has methods for working with events.
 type events struct {
-	eventSwitch *evts.EventSwitch
+	eventSwitch *go_events.EventSwitch
 }
 
 // Subscribe to an event.
-func (this *events) Subscribe(subId, event string, callback func(evts.EventData)) error {
-	this.eventSwitch.AddListenerForEvent(subId, event, callback)
+func (this *events) Subscribe(subId, event string,
+	callback func(txs.EventData)) error {
+	cb := func(evt go_events.EventData) {
+		eventData, err := mapToOurEventData(evt)
+		if err != nil {
+			panic(fmt.Sprintf("Failed to map go-events EventData to our EventData %v",
+				err))
+		}
+		callback(eventData)
+	}
+	this.eventSwitch.AddListenerForEvent(subId, event, cb)
 	return nil
 }
 
@@ -67,7 +83,8 @@ type multiplexedEvents struct {
 }
 
 // Subscribe to an event.
-func (multiEvents *multiplexedEvents) Subscribe(subId, event string, callback func(evts.EventData)) error {
+func (multiEvents *multiplexedEvents) Subscribe(subId, event string,
+	callback func(txs.EventData)) error {
 	for _, eventEmitter := range multiEvents.eventEmitters {
 		err := eventEmitter.Subscribe(subId, event, callback)
 		if err != nil {
@@ -118,5 +135,25 @@ func GenerateSubId() (string, error) {
 	}
 	rStr := hex.EncodeToString(b)
 	return strings.ToUpper(rStr), nil
+}
 
+func mapToOurEventData(eventData anyEventData) (txs.EventData, error) {
+	// While we depend on go-events in the way we do, we don't have much choice
+	// than to use a generic interface like anyEventData with a type switch.
+
+	switch eventData := eventData.(type) {
+	case txs.EventData:
+		return eventData, nil
+	case tm_types.EventDataNewBlock:
+		return txs.EventDataNewBlock{
+			Block: eventData.Block,
+		}, nil
+	case tm_types.EventDataNewBlockHeader:
+		return txs.EventDataNewBlockHeader{
+			Header: eventData.Header,
+		}, nil
+	default:
+		return nil, fmt.Errorf("EventData not recognised as known EventData: %v",
+			eventData)
+	}
 }
--- a/event/events.go
+++ b/event/events.go
@@ -24,16 +24,18 @@ import (
 	"fmt"
 
 	"github.com/eris-ltd/eris-db/txs"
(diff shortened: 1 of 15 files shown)


event/events.go
```diff
@@ -47,12 +54,21 @@ func Multiplex(events ...EventEmitter) *multiplexedEvents {
...
 // Subscribe to an event.
-func (this *events) Subscribe(subId, event string, callback func(evts.EventData)) error {
-	this.eventSwitch.AddListenerForEvent(subId, event, callback)
+func (this *events) Subscribe(subId, event string,
+	callback func(txs.EventData)) error {
+	cb := func(evt go_events.EventData) {
+		eventData, err := mapToOurEventData(evt)
+		if err != nil {
+			panic(fmt.Sprintf("Failed to map go-events EventData to our EventData %v",
+				err))
```
silasdavis: This probably rings alarm bells, but I don't have anywhere sensible to send an Error here. Really it should happen if we have an up-to-date mapping against the tendermint version we are using.

Maybe it's better to drop events on the floor that we don't understand?

event/events.go
```diff
@@ -47,12 +54,21 @@ func Multiplex(events ...EventEmitter) *multiplexedEvents {
...
 
 // Subscribe to an event.
-func (this *events) Subscribe(subId, event string, callback func(evts.EventData)) error {
-	this.eventSwitch.AddListenerForEvent(subId, event, callback)
+func (this *events) Subscribe(subId, event string,
+	callback func(txs.EventData)) error {
+	cb := func(evt go_events.EventData) {
+		eventData, err := mapToOurEventData(evt)
+		if err != nil {
+			panic(fmt.Sprintf("Failed to map go-events EventData to our EventData %v",
```
benjaminbollen: We cannot introduce new panics at this point; an error is sufficient.

event/events.go
```diff
@@ -21,20 +21,27 @@ import (
...
+	go_events "github.com/tendermint/go-events"
+	tm_types "github.com/tendermint/tendermint/types"
 )
 
 // TODO improve
 // TODO: [ben] yes please ^^^
 // [ben] To improve this we will switch out go-events with eris-db/event so
 // that there is no need anymore for this poor wrapper.
 
+// Oh for a sum type
```
benjaminbollen: a void pointer `anyEventData` without argumentation seems uncalled for.

event/events.go
benjaminbollen: I had not refreshed the page, so you saw my comment coming :). It is a difficult point; I would look at having the `callback` function accept the error as an argument; If that causes too much work just now; I would prefer logging, dropping failed event and adding a clear TODO comment.

event/events.go
```diff
@@ -118,5 +135,25 @@ func GenerateSubId() (string, error) {
 	}
 	rStr := hex.EncodeToString(b)
 	return strings.ToUpper(rStr), nil
+}
 
+func mapToOurEventData(eventData anyEventData) (txs.EventData, error) {
+	// While we depend on go-events in the way we do, we don't have much choice
```
benjaminbollen: Can we not define a new `txs` event type (or even define it at `/consensus`) that wraps around the go-events such that we swallow these event types, rather than this naked switch?  -- I'm not radically opposed if this resolves the bug in the short-term and we can add a comment of following discussion.

rpc/tendermint/test/common.go
```diff
@@ -32,7 +32,7 @@ func TestWrapper(runner func() int) int {
 func DebugMain() {
 	t := &testing.T{}
 	TestWrapper(func() int {
-		testBroadcastTx(t, "HTTP")
```
benjaminbollen: What happened here?  -- looks like a non-trivial delta out of context, ie I don't get it :)

event/events.go
silasdavis: I considered something like this, but since go-events doesn't use an erring callback it gets a bit weird with asynchrony. If you really wanted to do it I think using some error channels that you then need to check would be the way. I think we're trying to guard against a 'our coding' problem here rather than something we should really expect at runtime (we control the tendermint version).

Perhaps a nicer middle ground is to `log.Error(...)`. That way it should show up as serious but the whole stack doesn't go down just because someone subscribes to a tendermint event we don't understand..?

event/events.go
silasdavis: So I think this is addressed in previous comment. I don't have a nice option here for returning an error along the callback chain.

I am leaning towards `log.Error`ing

event/events.go
silasdavis: I need to accept `interface{}` because that is what a `TMEventData` (https://github.com/tendermint/tendermint/blob/master/types/events.go#L36), via go-events `EventData` (https://github.com/tendermint/go-events/blob/master/events.go#L11).

So creating a named type that is an `interface{}` is just an idiom that documents what I am expecting here.

rpc/tendermint/test/common.go
silasdavis: DebugMain is my convenience method that allows me to run delve, it's not part of normal testing.

event/events.go
silasdavis: I'm really not sure there is a nicer way to do this. In essence `anyEventData` is wrapping around go-events `EventData`. I could embed `go_events.EventData` in `anyEventData`, but bear in mind that is just `interface{}` so the effect would be the same.

If we have decided that we want to have these event types in our code (I think we have) then this allows us to do more complex mapping/normalisation of events if we ever need to, for example for other consensus engines. But using a type switch like this is regrettable how this sort of thing is done in go, I think.

I'm open to suggestions if you know of a better pattern or have something specific I haven't thought of when you say 'wraps around go-events'

rpc/tendermint/test/common.go
silasdavis: "Non-trivial delta out of context"... I'm just going to leave this [here](http://www.plainenglish.co.uk/)

event/events.go
silasdavis: Also 'void pointer'? Have you been secretly writing C?

rpc/tendermint/test/common.go
benjaminbollen: No, I think putting a short comment above the function is more than adequate to argue for it's existence (referencing delve as a debugging tool)

event/events.go
benjaminbollen: Well, I fully agree this is a problematic construction and we need to sort it out first.  But this is why I committed a proposal base code (copied from go-ethereum) in `/event_new`; because the only solution I see is building this event stream from the ground up.  In my opinion, you're right to address this like this, but it's a temporary solution.

event/events.go
benjaminbollen: OK. Mixing async paradigms (error channel where it's currently using callbacks) is not a net-gain on clarity, so let's not go there.  Im good with logging it and dropping it.

benjaminbollen: But do add a TODO please

event/events.go
silasdavis: OIC right yeah temporary WRT that I agree

event/events.go
silasdavis: TODO: use a language where not closing over a sum type is a compile-time error.

event/events.go
```diff
@@ -21,20 +21,29 @@ import (
...
-// TODO: [ben] yes please ^^^
-// [ben] To improve this we will switch out go-events with eris-db/event so
-// that there is no need anymore for this poor wrapper.
+// TODO: [Silas] this is a compatibility layer between our event types and
+// TODO: go-events. Our ultimate plan is to replace go-events with our own pub-sub
+// TODO: code that will better allow us to manage and multiplex events from different
+// TODO: subsystems
+
+// Oh for a sum type
+// We are using this as a marker interface for the
```
benjaminbollen: for the ... ?

silasdavis: woops, I'll fix that

PR #226, opened 2016-08-18
Stop tests from panicking and fix stale cache issue

silasdavis:
Some changes here are to make the tests not panic. There is also in eris-mint.go a change to remove a stale cache bug.

### DIFF
--- a/manager/eris-mint/state/execution.go
+++ b/manager/eris-mint/state/execution.go
@@ -279,6 +279,7 @@ func adjustByInputs(accounts map[string]*acm.Account, ins []*txs.TxInput) {
 			PanicSanity("adjustByInputs() expects sufficient funds")
 		}
 		acc.Balance -= in.Amount
+
 		acc.Sequence += 1
 	}
 }
@@ -413,6 +414,7 @@ func ExecTx(blockCache *BlockCache, tx txs.Tx, runCall bool, evc events.Fireable
 
 		// Good!
 		value := tx.Input.Amount - tx.Fee
+
 		inAcc.Sequence += 1
 		inAcc.Balance -= tx.Fee
 		blockCache.UpdateAccount(inAcc)
--- a/rpc/tendermint/core/routes.go
+++ b/rpc/tendermint/core/routes.go
@@ -55,6 +55,12 @@ func (tmRoutes *TendermintRoutes) GetRoutes() map[string]*rpc.RPCFunc {
 func (tmRoutes *TendermintRoutes) Subscribe(wsCtx rpctypes.WSRPCContext,
 	event string) (ctypes.ErisDBResult, error) {
 	// NOTE: RPCResponses of subscribed events have id suffix "#event"
+	// TODO: we really ought to allow multiple subscriptions from the same client address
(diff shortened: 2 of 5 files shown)


manager/eris-mint/eris-mint.go
```diff
@@ -185,15 +185,14 @@ func (app *ErisMint) Commit() (res tmsp.Result) {
...
-	// if there were any txs in the block,
-	// reset the check cache to the new height
-	if app.nTxs > 0 {
-		log.WithFields(log.Fields{
-			"txs": app.nTxs,
-		}).Info("Reset checkCache")
-		app.checkCache = sm.NewBlockCache(app.state)
-	}
-	app.nTxs = 0
+	// Refresh checkCache even if there were no transactions. A CheckTx may have
```
benjaminbollen: The concurrency model of TMSP application interface is still troublesome and we need to enforce it from the application side.  However, is the mempool not supposedly locked during commit, such that CheckTx and AppendTx are not called?

Can we consider a lock controlled by commit on holding all CheckTx/AppendTx, and holding commit until all CheckTx and AppendTx have cleared out of the queue?  -- I ve held off on this because we need to revisit the async model here at a fundamental level.

I am concerned here that undoing the nTxs breaks the original assumptions in this code.

benjaminbollen: Mulling on it a bit more; checkCache is okay to be ahead of the real application state; in addition to the above remark that we need to strongly enforce the concurrency from within erismint, CheckTx should be more strongly aware of possible of its inability to acknowledge the ordering and as such can be made smarter on explicit validations of ordering; eg sequence number on CheckTx should be a strict inequality, rather than exactly equal to `sequence_number + 1`

benjaminbollen:
@silasdavis can you git force this chance on eris-mint into a separate PR; then Im happy to review and merge the test fixes;  with regards to the erismint change, I want to hold a reservation; elaborate discussion and maybe we can even work on a unit test

benjaminbollen:
"Tests part" looks good to me

manager/eris-mint/eris-mint.go
silasdavis: > CheckTx can be made more strongly aware of its inability to acknowledge the ordering and as such can be made smarter on explicit validations of ordering;
I'm not entirely sure what you are are saying in your second comment

I don't really understand what you are saying here. The ordering of any previous block's transaction are obviously fixed

When we commit we want to make sure future CheckTx runs have a fresh state, checking here for whether there is an empty block is a completely premature optimisation. I think it gains us very little and makes the code harder to understand.

The issue is that CheckTx can get run once and before being added to a proposal a recheck is forced (by a new block), the recheck should be running against the last committed state not the dirty state from the last CheckTx. Strict inequality wouldn't help here, I have a Tx with a correct nonce, but it is deemed invalid because it's been applied to the state twice. Actually it's valid against the current committed state.

Come to think of it, we should really be refreshing the checkCache every time CheckTx is called! I will remove this from here and make an analternative proposal we can discuss.

benjaminbollen:
LGTM, or FIPF

Issue #228, opened 2016-08-19
[Pipe] ErisMint: checkCache for CheckTx reset


benjaminbollen:
Currently if an empty block is committed (no transactions included) then the cached copy for CheckTx - which runs in parallel to AppendTx - is not reset and preserved.  The checkCache is reset to the current state when a non-empty block is committed.

Tendermint mempool however can update and recheck the transactions; this can make the checkCache and state move out of sync;
1. One consideration is that ErisMint needs to synchronise on FlushSync() 
2. A tension remains: on a new block, the application would want to reset caches as the block tick is the only reference point; on the other hand, one can argue that the mempool already improves statistically the odds of producing a synchronized sequence of transactions, if indeed the checkCache is preserved (allowing sequentially dependent transactions to potentially be included in the same block*); It has to be added though that if no transactions are committed in the block, then it is likely to assume there are close to no transactions lingering in the mempool.  This provides the strongest argument to stay close to the logic of preserving the block tick for resetting caches.  -> This resolves the tension lined out in point 2., but builds a tension with point 1.

*This can be understood by considering that if CheckTx would work off off the last committed state for every checked transaction, the sequence numbers of sequential transactions in the same block would cause a dependent transaction to be rejected from the mempool.

silasdavis:
Great this capture what we spoke about perfectly.

compleatang:
Handled in new concurrency paradigm.

PR #229, opened 2016-08-20
Don't special case empty blocks when refreshing the checkCache on commit

silasdavis:
There are more profound related considerations detailed in #228. But independent of those I think this simple change is a win for clarity and the principle of honouring the most recent committed state.

Only refreshing the checkCache on non-empty blocks can be seen as an optimisation that it turns out is on shaking grounds: namely assuming that the committing of an empty block implies that the checkCache is clean. This may have been true at one point, but no longer is in general.

This change also fixes an actual real life bug in the reprocessing of an identical transaction in the mempool that is encountered in integration tests.

### DIFF
--- a/manager/eris-mint/eris-mint.go
+++ b/manager/eris-mint/eris-mint.go
@@ -185,14 +185,11 @@ func (app *ErisMint) Commit() (res tmsp.Result) {
 	// sync the AppendTx cache
 	app.cache.Sync()
 
-	// if there were any txs in the block,
-	// reset the check cache to the new height
-	if app.nTxs > 0 {
-		log.WithFields(log.Fields{
-			"txs": app.nTxs,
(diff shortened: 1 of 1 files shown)


benjaminbollen:
Discussion can be had on issue #228

silasdavis:
I'd like to separately merge this, and have the more general solution discussed there

benjaminbollen:
Given the design of Tendermint Socket Protocol, we need to raise an issue and harmonise this with FlushSync (and FlushAsync); but in general I also do not see that this introduces an error in the logic (as I initially was concerned about in your previous PR), and for summary of arguments see #228.

I think you are right that we should drive the logic as close as possible to the blocktick.

benjaminbollen:
Looks good to me, but we should keep this change in the back of our heads as we evaluate it in depth.

Issue #230, opened 2016-08-21
Implement missing state and functions on erisMintPipe that need to call Tendermint's


silasdavis:
`erisMintPipe` contains three pieces of state that depend on calling out to Tendermint's RPC (for us in-process) and none of them are fully implemented, see: https://github.com/eris-ltd/eris-db/blob/develop/manager/eris-mint/pipe.go#L115 

The fields on `erisMintPipe` storing these pieces of state are:

1. `network`
2. `consensus`
3. `blockchain`

There are various `\\ TODO: RPC` comments in bits of their implementation, and blockchain and consensus are currently not being intialised at all.

We may be able to get away with not implementing all of them at this stage.

PR #231, opened 2016-08-21
Tweak BlockchainInfo to use stuff we have actually initialised

silasdavis:
This re-implements BlockchainInfo to work around #230. This implementation actually works, but unit tests will follow with #218, which require gently mocking erisMintPipe behind its back.


benjaminbollen:
A mock of ErisMintPipe exists in `/tests/mock`

benjaminbollen:
LGTM; I would even phrase it stronger than you have in the issue description: the correct implementation of `BlockchainInfo` is over the `consensusEngine` interface.  The unit tests should be both purely at RPC level with a hollow mock of Pipe; and at the consensusEngine interface, with a mock of consensusEngine

benjaminbollen:
I noticed that the tests of Event are broken; this is - by assumption - unrelated to the current PR, and has been merged under the radar in a previous PR when touching on the events structure.

PR #232, opened 2016-08-21
Dry up client

silasdavis:
wet dog

### DIFF
--- a/rpc/tendermint/client/client.go
+++ b/rpc/tendermint/client/client.go
@@ -1,6 +1,7 @@
 package client
 
 import (
+	"fmt"
 	acm "github.com/eris-ltd/eris-db/account"
 	core_types "github.com/eris-ltd/eris-db/core/types"
 	rpc_types "github.com/eris-ltd/eris-db/rpc/tendermint/core/types"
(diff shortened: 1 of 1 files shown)


rpc/tendermint/client/client.go
```diff
@@ -159,36 +94,68 @@ func CallCode(client rpcclient.Client,
...
+		_, err = cli.Call(method, paramsSlice, &res)
+	case *rpcclient.ClientURI:
+		_, err = cli.Call(method, paramsMap, &res)
+	}
+	return res, err
+
+}
+func mapAndValues(orderedKeyVals ...interface{}) (map[string]interface{}, []interface{}) {
+	if len(orderedKeyVals)%2 != 0 {
+		panic(fmt.Sprintf("mapAndValues requires a even length list of keys and "+
```
benjaminbollen: These panics should not be panics, rather errors.  I can see why you use a panic here; but I would only want to move back to panics here, if we had asserted coverage of all calls to `mapAndValues` in the CI; right now this relies on our continued vigilance.

silasdavis: Yeah I guess I could have just let it panic array out of bounds, so I saw this as making the panic nicer. But I agree.

silasdavis:
@benjaminbollen Stopped panicking

benjaminbollen:
LGTM

PR #235, opened 2016-08-22
Use a random subscription id as rpc/v0 does on rpc/tendermint

### LINKED ISSUE #172: Implementing subscribe/unsubscribe against tendermint RPC
Implementation of these moved to tendermint:

https://github.com/tendermint/tendermint/blob/master/rpc/core/routes.go#L11

And:

https://github.com/tendermint/tendermint/blob/master/rpc/core/events.go#L10

Should be portable

### LINKED ISSUE #225: rpc/v/tendermint Subscribe needs to allow multiple connections from same host
Currently the implementation only allows one subscriptions for the same event from the same host. This is robust and is not assumed by existing client libraries. v0 API does something sensible by issuing a random subscription id. should use that.

silasdavis:
Fixes #177.
Fixes #172.
Fixes #225.

### DIFF
--- a/rpc/tendermint/test/tests.go
+++ b/rpc/tendermint/test/tests.go
@@ -6,9 +6,12 @@ import (
 	"testing"
 
 	edbcli "github.com/eris-ltd/eris-db/rpc/tendermint/client"
+	core_types "github.com/eris-ltd/eris-db/rpc/tendermint/core/types"
 	"github.com/eris-ltd/eris-db/txs"
-
 	"github.com/stretchr/testify/assert"
+
+	"time"
+
 	tm_common "github.com/tendermint/go-common"
 	"golang.org/x/crypto/ripemd160"
 )
@@ -207,7 +210,7 @@ func testNameReg(t *testing.T, typ string) {
 	assert.Equal(t, user[0].Address, entry.Owner)
 
 	// update the data as the owner, make sure still there
-	numDesiredBlocks = int64(4)
+	numDesiredBlocks = int64(3)
 	const updatedData = "these are amongst the things I wish to bestow upon the youth of generations come: a safe supply of honey, and a better money. For what else shall they need"
 	amt = fee + numDesiredBlocks*txs.NameByteCostMultiplier*txs.NameBlockCostMultiplier*txs.NameBaseCost(name, updatedData)
 	tx = makeDefaultNameTx(t, typ, name, updatedData, amt, fee)
@@ -218,7 +221,6 @@ func testNameReg(t *testing.T, typ string) {
 	assert.Equal(t, updatedData, entry.Data)
 
 	// try to update as non owner, should fail
-	//waitBlocks(t, wsc, 4)
 	tx = txs.NewNameTxWithNonce(user[1].PubKey, name, "never mind", amt, fee,
 		getNonce(t, typ, user[1].Address)+1)
 	tx.Sign(chainID, user[1])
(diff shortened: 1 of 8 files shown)


benjaminbollen:
LGTM, but breaks API to clients (e-pm) and eris-db.js;  this is okay as 0.12.0 is a big transfer, but we need to coordinate.  Hence Im holding off merging and will ping @dennismckinnon @NodeGuy

silasdavis:
Contrary to what I had thought, 46657 did have a subscribe and unsubscribe websocket implementation here: https://github.com/eris-ltd/eris-db/blob/master/Godeps/_workspace/src/github.com/tendermint/tendermint/rpc/server/handlers.go#L331

This implementation only allows a single subscription to the same event from the same host. On the face of it that might seem okay, but it puts the onus on the client to maintain a global websocket connection pool per event and reuse/share the same one for the same event. This isn't terribly nice for writing less stateful Javascript apps and furthermore there is no warning that by re-registering the event will clobber any previous connections (they will no longer receive the event). This is because AddListener in go-events overwrites any existing callback:

```
func (cell *eventCell) AddListener(listenerID string, cb eventCallback) {
	cell.mtx.Lock()
	cell.listeners[listenerID] = cb
	cell.mtx.Unlock()
}
```

benjaminbollen:
Your arguments are very strong; I will want to wait for develop to be green; then rerun this build and then Im happy to push forward on this.

silasdavis:
I'm a bit worried about breaking compatibility, but using this approach the worst that happens is we break unsubscribe (they pass event name rather than subscription id, it's a no-op) -- something we have to rely on the clients doing anyway. Subscribe will work as before. And both methods have identical signatures so Unsubscribe will fail silently -- not the end of the world.

silasdavis:
Actually I guess the JSON versions would break because the key name has changed from `event` to `subscriptionId`. The HTTP version will fail silently.

silasdavis:
Rebased

rpc/tendermint/test/tests.go
```diff
@@ -251,7 +253,59 @@ func asEventDataTx(t *testing.T, eventData txs.EventData) txs.EventDataTx {
 	return eventDataTx
 }
 
-func doNothing(eventId string, eventData txs.EventData) (bool, error) {
+func doNothing(_ string, _ txs.EventData) (bool, error) {
 	// And ask waitForEvent to stop waiting
 	return true, nil
 }
+
+func testSubscribe(t *testing.T) {
```
benjaminbollen: Is this being run? as it is not exported?

silasdavis: Note this is not a *_test.go file.

It's following the same pattern as other methods in here, it's the testing method that the real one in https://github.com/silasdavis/eris-db/blob/f632852bb6e032292b9cbadb427b03c807ddb018/rpc/tendermint/test/client_ws_test.go#L241 calls.

Aside from following that pattern, the may reason I have it here is because go-wire doesn't work from *_test.go files it would seem. Something to do with the global var interface registration.

silasdavis: Oh actually since I factored out `readResult`. I could probably get away with having it in client_ws_test.go. Still I'm fairly happy having it here since all the tests in client_ws_test.go are proxy functions to ones here. Mind you they need to be to centralise the HTTP/JSON logic.

benjaminbollen: Points taken onboard.

benjaminbollen:
LGTM

NodeGuy:
I don't understand this conversation.

I assume that we're using [JSON-RPC](http://json-rpc.org/) as the foundation for our protocol.  Currently the JavaScript libraries seem to implement their own version of JSON-RPC instead of using a library!  I would like to be able to drop in a library such as the [Multitransport JSON-RPC Client](https://github.com/uber/multitransport-jsonrpc) and have it **Just Work** over HTTP and WebSocket.

Can I do that with the change you're proposing?

PR #237, opened 2016-08-22
Fix mapAndValues and even write some, like, test

silasdavis:


rpc/tendermint/client/client.go
```diff
@@ -158,7 +159,7 @@ func mapAndValues(orderedKeyVals ...interface{}) (map[string]interface{},
 		}
 		val := orderedKeyVals[i+1]
 		paramsMap[key] = val
-		paramsSlice = append(paramsSlice, val)
+		paramsSlice[i/2] = val
```
benjaminbollen: I dislike golang for allowing this.

benjaminbollen:
LGTM

rpc/tendermint/client/client.go
benjaminbollen: Rather I should say, I'd prefer we multiply *2 in line 154; over dividing here; but fine :)

PR #240, opened 2016-08-23
version: correct develop to v0.12.0-rc3

benjaminbollen:
- correct version on develop to v0.12.0-rc3

- improve the assertion of modules (specifically consensus/tendermint) to read semantic version of dependency from vendored source code, and assert whether it is compatible with the expected version.

- also addresses deprecation item in https://github.com/eris-ltd/eris-db/issues/236

### DIFF
--- a/consensus/tendermint/version.go
+++ b/consensus/tendermint/version.go
@@ -17,6 +17,10 @@
 package tendermint
 
 import (
+	"strconv"
+
+	tendermint_version "github.com/tendermint/tendermint/version"
+
 	version "github.com/eris-ltd/eris-db/version"
 )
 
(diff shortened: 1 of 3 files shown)


consensus/tendermint/version.go
```diff
@@ -17,21 +17,62 @@
...
 	// Minor version component of the current release
-	tendermintVersionMinor = 6
+	tendermintVersionMinorConst uint8 = 6
 	// Patch version component of the current release
-	tendermintVersionPatch = 0
+	tendermintVersionPatchConst uint8 = 0
+)
+
+var (
+	tendermintVersionMajor uint8
```
silasdavis: What about version 256...?

silasdavis:
Ok nice this makes loads of sense now I catch up

PR #243, opened 2016-08-24
Update opcodes.go

VoR0220:
* includes delegatecall opcode. May need to include necessary changes in other files but this is a critical component for library usage and needs to definitely be implemented in the evm.

### DIFF
--- a/manager/eris-mint/evm/opcodes.go
+++ b/manager/eris-mint/evm/opcodes.go
@@ -167,6 +167,7 @@ const (
 	CALL
 	CALLCODE
 	RETURN
+	DELEGATECALL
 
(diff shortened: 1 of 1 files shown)


benjaminbollen:
Can you link to the eth documentation, because the bytecode needs to match and an implementation will be good to have

benjaminbollen:
first addressing this in #247 on old tendermint

benjaminbollen:
Thanks; I'll leave the branch, as for implementing `delegatecall` on `develop` / e-db 0.12.0-rc3 I will want to pull in your start-off commit.  Only for edb on `master` I had to do it in the godeps

PR #247, opened 2016-08-26
[REVIEW] issue244 delegatecall

benjaminbollen:
this is for communication, review and testing purposes only.  Do not merge this into master, until we have properly reviewed and tested

### DIFF
--- a/Godeps/_workspace/src/github.com/tendermint/tendermint/vm/vm.go
+++ b/Godeps/_workspace/src/github.com/tendermint/tendermint/vm/vm.go
@@ -147,18 +147,16 @@ func (vm *VM) Call(caller, callee *Account, code, input []byte, value int64, gas
 	return
 }
 
-
-// [This implementation is up for review].
 // DelegateCall is executed by the DELEGATECALL opcode, introduced as off Ethereum Homestead.
 // The intent of delegate call is to run the code of the callee in the storage context of the caller;
 // while preserving the original caller to the previous callee.
-// [ -->] Different to the normal CALL or CALLCODE, the value does not need to be transferred to the callee. [<-- CORRECT?]
+// Different to the normal CALL or CALLCODE, the value does not need to be transferred to the callee.
 func (vm *VM) DelegateCall(caller, callee *Account, code, input []byte, value int64, gas *int64) (output []byte, err error) {
-	
+
 	exception := new(string)
 	// fire the post call event (including exception if applicable)
 	defer vm.fireCallEvent(exception, &output, caller, callee, input, value, gas)
-	
+
 	// DelegateCall does not transfer the value to the callee.
 
 	if len(code) > 0 {
@@ -781,11 +779,11 @@ func (vm *VM) call(caller, callee *Account, code, input []byte, value int64, gas
 			// caller, as such it is not stored on stack for the address
 			// to be called; for CALL and CALLCODE value is stored on stack
 			// and needs to be overwritten from the given value.
-			// TODO: [ben] assert that malicious code cannot produce 
+			// TODO: [ben] assert that malicious code cannot produce
 			// new value.
 			if op != DELEGATECALL {
 				value = stack.Pop64()
-			} 
+			}
 			inOffset, inSize := stack.Pop64(), stack.Pop64()   // inputs
 			retOffset, retSize := stack.Pop64(), stack.Pop64() // outputs
 			dbg.Printf(" => %X\n", addr)
--- a/Godeps/_workspace/src/github.com/tendermint/tendermint/vm/vm.go
+++ b/Godeps/_workspace/src/github.com/tendermint/tendermint/vm/vm.go
@@ -771,7 +771,7 @@ func (vm *VM) call(caller, callee *Account, code, input []byte, value int64, gas
 				stack.Push(newAccount.Address)
 			}
 
-		case CALL, CALLCODE: // 0xF1, 0xF2
+		case CALL, CALLCODE, DELEGATECALL: // 0xF1, 0xF2, 0xF4
 			if !HasPermission(vm.appState, callee, ptypes.Call) {
 				return nil, ErrPermission{"call"}
 			}
--- a/Godeps/_workspace/src/github.com/tendermint/tendermint/vm/vm.go
+++ b/Godeps/_workspace/src/github.com/tendermint/tendermint/vm/vm.go
@@ -147,6 +147,32 @@ func (vm *VM) Call(caller, callee *Account, code, input []byte, value int64, gas
 	return
 }
 
+
+// [This implementation is up for review].
+// DelegateCall is executed by the DELEGATECALL opcode, introduced as off Ethereum Homestead.
+// The intent of delegate call is to run the code of the callee in the storage context of the caller;
+// while preserving the original caller to the previous callee.
+// [ -->] Different to the normal CALL or CALLCODE, the value does not need to be transferred to the callee. [<-- CORRECT?]
+func (vm *VM) DelegateCall(caller, callee *Account, code, input []byte, value int64, gas *int64) (output []byte, err error) {
+	
+	exception := new(string)
+	// fire the post call event (including exception if applicable)
+	defer vm.fireCallEvent(exception, &output, caller, callee, input, value, gas)
+	
+	// DelegateCall does not transfer the value to the callee.
+
+	if len(code) > 0 {
+		vm.callDepth += 1
+		output, err = vm.call(caller, callee, code, input, value, gas)
+		vm.callDepth -= 1
+		if err != nil {
+			*exception = err.Error()
+		}
+	}
+
+	return
+}
+
 // Try to deduct gasToUse from gasLeft.  If ok return false, otherwise
 // set err and return true.
 func useGasNegative(gasLeft *int64, gasToUse int64, err *error) bool {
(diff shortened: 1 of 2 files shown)


benjaminbollen:
so reasoning is:
- act on storage of `callee`
- preserve `caller`
- run code from destination
- do not transfer value to destination

VoR0220:
Did you look at this yet? https://github.com/ethereum/go-ethereum/search?utf8=%E2%9C%93&q=delegatecall This has everywhere that is delegate call related in some way and could be an excellent reference point.

Godeps/_workspace/src/github.com/tendermint/tendermint/vm/vm.go
```diff
@@ -147,6 +147,32 @@ func (vm *VM) Call(caller, callee *Account, code, input []byte, value int64, gas
 	return
 }
 
+
+// [This implementation is up for review].
+// DelegateCall is executed by the DELEGATECALL opcode, introduced as off Ethereum Homestead.
+// The intent of delegate call is to run the code of the callee in the storage context of the caller;
+// while preserving the original caller to the previous callee.
+// [ -->] Different to the normal CALL or CALLCODE, the value does not need to be transferred to the callee. [<-- CORRECT?]
```
VoR0220: Not entirely correct. See here. https://github.com/ethereum/go-ethereum/blob/b7e3dfc5a2bc7e2f4d653fbe0ec9774277a10643/core/execution.go#L41

silasdavis:
I'm going to state my understanding of this operation and see if other people agree. "Caller" and "Callee" can get a bit confusing because we an intermediate account (named B below; the delegate) that is both caller and callee.

Suppose we have 3 accounts: A, B, and C with the following psuedo-code contracts attached:

A: 
```
int n := 0
address addr := nil

CALL B
```
B: 
```
int n := 0
address addr := nil

DELEGATECALL C
```
C:
```
int n := 0
address addr := nil

n++
addr = msg.sender
```

Then the result of invoking A (message or transaction) should be the following sequence:

| Stack | A | B | C |
|---------|----|----|----|
| []  | n = 0, addr = nil  | n = 0, addr = nil  | n = 0, addr = nil  |
| [CALL B]  | n = 0, addr = nil  | n = 0, addr = nil  | n = 0, addr = nil  |
| [DELEGATECALL C, CALL B]  | n = 0, addr =nil  | n = 1, addr = A  | n = 0, addr = nil  |

That is B acts a 'delegate' for A in terms of the message sender and value, but like with `CALLCODE` it is B's storage that gets updated by the code in C.

Is this correct?

silasdavis:
I need clarification on what this means, and if it is indeed correct...

>`// [ -->] Different to the normal CALL or CALLCODE, the value does not need to be transferred to the callee. [<-- CORRECT?]`

benjaminbollen:
> That is B acts a 'delegate' for A in terms of the message sender and value, but like with CALLCODE it is > B's storage that gets updated by the code in C.
> Is this correct?
yes, that is also how i understand it and the pseudo-code example is consistent.

benjaminbollen:
> // [ -->] Different to the normal CALL or CALLCODE, the value does not need to be transferred to the > callee. [<-- CORRECT?]
> I need clarification on what this means, and if it is indeed correct...

in the code, opcodes `CALL` and `CALLCODE` will start a new stack and memory by calling the funnction `Call`, which atomically transfers the value from caller to callee; then calls the function `call` which executes the ethereum opcodes;

I understand that the value sent with the `message` (starting a new stack+memory) does not require the value sent from `A` to `B` to be passed to the delegated contract `C`; a few notes:
- whether or not to pass on this value from B to C; does not affect the succes of e-pm test `app35`
- go-ethereum code - from my reading - does also not pass this value to the delegated contract `C`

compleatang:
the key diff between CALL and DELEGATECALL is the storage and memory zone in which the code operates. namely CALL will execute in the storage and memory zone of the CALLEE while DELEGATECALL will execute the in the storage and memory zone of the CALLER.

benjaminbollen:
so in the proposed impl here; for `switch CALL, CALLCODE, DELEGATECALL` it essential hinges on this: 
```
if op == CALLCODE {
    [...]
    ret, err = vm.Call(callee, callee, acc.Code, args, value, gas)
} else if op == DELEGATECALL {
    [...]				
    ret, err = vm.DelegateCall(caller, callee, acc.Code, args, value, gas)
} else {
    [...]
    ret, err = vm.Call(callee, acc, acc.Code, args, value, gas)
}
```
https://github.com/eris-ltd/eris-db/pull/247/files#diff-cad6213f6a08c87e4110c4006789e2b7R832

compleatang:
I'm pretty sure there's a few nuances in the proposed `vm.DelegateCall` which are not captured in geths -> https://github.com/ethereum/go-ethereum/blob/b7e3dfc5a2bc7e2f4d653fbe0ec9774277a10643/core/execution.go#L138-L169

VoR0220:
The original EIP with all the specifications. https://github.com/ethereum/EIPs/issues/23

silasdavis:
I just want to emphasise the difference between `DELEGATECALL` and `CALLCODE` in the example I gave before. I think the following execution would occur with `CALLCODE`: 

| Stack | A | B | C |
|---------|----|----|----|
| []  | n = 0, addr = nil  | n = 0, addr = nil  | n = 0, addr = nil  |
| [CALL B]  | n = 0, addr = nil  | n = 0, addr = nil  | n = 0, addr = nil  |
| [CALLCODE C, CALL B]  | n = 0, addr = nil  | n = 1, addr = **B**  | n = 0, addr = nil  |

Do we agree?

benjaminbollen:
Yes, I agree that the stored address would be now stored as `B` as the message sender.

version/version.go
```diff
@@ -19,4 +19,4 @@ package version
 const TENDERMINT_VERSION = "0.5.0"
 // IMPORTANT: Eris-DB version must be on the last line of this file for
 // the deployment script DOCKER/build.sh to pick up the right label.
-const VERSION = "0.12.0"
+const VERSION = "0.12.0-rc2"
```
pietv: @benjaminbollen it's "0.12.0" for #master. without rc.

Godeps/_workspace/src/github.com/tendermint/tendermint/vm/vm.go
VoR0220: Just change these notes here please @benjaminbollen then it should be ready to merge.

Godeps/_workspace/src/github.com/tendermint/tendermint/vm/vm.go
```diff
@@ -745,12 +769,21 @@ func (vm *VM) call(caller, callee *Account, code, input []byte, value int64, gas
...
 				return nil, ErrPermission{"call"}
 			}
 			gasLimit := stack.Pop64()
-			addr, value := stack.Pop(), stack.Pop64()
+			addr := stack.Pop()
+			// NOTE: for DELEGATECALL value is preserved from the original
+			// caller, as such it is not stored on stack for the address
+			// to be called; for CALL and CALLCODE value is stored on stack
+			// and needs to be overwritten from the given value.
+			// TODO: [ben] assert that malicious code cannot produce
```
silasdavis: We keeping this as todo?

benjaminbollen: It is arguably redundant as value cannot be passed to another contract if it did not exist; and value is atomically swapped.  However, should a contract rely on `CALLVALUE` Im at this point not confident it would not be possible to produce bytecode that would put a malignent value on the stack and as such provide a hidden way to corrupt seemingly valid solidity.

Godeps/_workspace/src/github.com/tendermint/tendermint/vm/vm.go
```diff
@@ -745,12 +769,21 @@ func (vm *VM) call(caller, callee *Account, code, input []byte, value int64, gas
...
 			if !HasPermission(vm.appState, callee, ptypes.Call) {
 				return nil, ErrPermission{"call"}
 			}
 			gasLimit := stack.Pop64()
-			addr, value := stack.Pop(), stack.Pop64()
+			addr := stack.Pop()
+			// NOTE: for DELEGATECALL value is preserved from the original
+			// caller, as such it is not stored on stack for the address
+			// to be called; for CALL and CALLCODE value is stored on stack
+			// and needs to be overwritten from the given value.
```
silasdavis: I found this a bit hard to parse, having read it a few times over. Here is a proposed rewording for clarity, but happy to see an even better version if this doesn't convey proper meaning:

```
			// NOTE: for DELEGATECALL value is preserved from the original
			// caller, as such it is not stored on stack for as an argument for
                        // DELEGATECALL and should not be popped. Instead the previous
                        // caller value is use.
                        // for CALL and CALLCODE value is stored on stack
			// and needs to overwrite the original caller value.
```

silasdavis:
So other than my suggestion for a rewording of the comment describing the crucial implementation fact for DELEGATECALL, this looks good to me!

Great work Ben

Godeps/_workspace/src/github.com/tendermint/tendermint/vm/vm.go
benjaminbollen: I was making these changes on develop; ok to just take the new comment on 0.12 develop? And leave it for hotfix here as is?

PR #250, opened 2016-09-01
significantly reduce build time of Dockerfile

compleatang:


Dockerfile
```diff
@@ -8,25 +8,20 @@ EXPOSE 46656
...
+# install eris-db's dependencies
 
 # set the source code path and copy the repository in
 ENV ERIS_DB_SRC_PATH $GOPATH/src/github.com/eris-ltd/eris-db
-COPY . $ERIS_DB_SRC_PATH
+ADD glide.yaml $ERIS_DB_SRC_PATH/
+ADD glide.lock $ERIS_DB_SRC_PATH/
+# [csk] if we vendor the dependencies we should import them b4 the glide install, no?
 
 # fetch and install eris-db and its dependencies
```
therealplato: dupe of L11/L47
```diff
- # fetch and install eris-db and its dependencies
```

compleatang: good catch! thanks.

silasdavis:
We also have: https://github.com/eris-ltd/eris-db/blob/develop/Dockerfile.armhf which should been shown the same love. I don't want these two Dockerfiles to get out of sync, if this become a burden we might consider scripting it:

https://github.com/devTristan/dockerscript
https://github.com/avirshup/DockerMake

There is some talk of a Dockerfile `INCLUDE` statement which would also help us not repeat ourselves in these two files.

For the time being could you make the corresponding changes in both places.

benjaminbollen:
smart!  happy to merge this and we will update the armhf dockerfile accordingly.

PR #256, opened 2016-09-05
Simulated call

VoR0220:
Attempts to run a simulated call whereby no actions affect the chain (my source for this is how they were done previously, but I am more than open to an alternative suggestion)

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -30,7 +30,7 @@ import (
 	"github.com/tendermint/go-crypto"
 	"github.com/tendermint/go-rpc/client"
 
-	// ptypes "github.com/eris-ltd/permission/types"
+	//ptypes "github.com/eris-ltd/permission/types"
 
 	log "github.com/eris-ltd/eris-logger"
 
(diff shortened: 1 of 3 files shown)


client/core/core.go
```diff
@@ -69,50 +69,74 @@ func Send(nodeAddr, signAddr, pubkey, addr, toAddr, amtS, nonceS string) (*txs.S
...
+	}
 
-// 	tx := types.NewCallTxWithNonce(pub, toAddrBytes, dataBytes, amt, gas, fee, int(nonce))
-// 	return tx, nil
-// }
+	tx := txs.NewCallTxWithNonce(pub, toAddrBytes, dataBytes, amt, gas, fee, int(nonce))
+	return tx, nil
+}
 
-// func Name(nodeAddr, signAddr, pubkey, addr, amtS, nonceS, feeS, name, data string) (*txs.NameTx, error) {
```
benjaminbollen: This module has a set of functions that each construct the correspondingly named transaction format.  Simulated call is already accessible through the RPC, we do not want to confuse that here.

VoR0220: yeah I messed that up here between all my git checkouts.

VoR0220: How do I access Simulated Call through the RPC?

benjaminbollen:
will track in #264

PR #257, opened 2016-09-05
Needed params from chain status

VoR0220:
Needed the latest block height, so I plopped that in there.

### DIFF
--- a/client/cmd/transaction.go
+++ b/client/cmd/transaction.go
@@ -66,7 +66,7 @@ func buildTransactionCommand() {
 	}
 	nameCmd.Flags().StringVarP(&clientDo.AmtFlag, "amt", "a", "", "specify an amount")
 	nameCmd.Flags().StringVarP(&clientDo.NameFlag, "name", "n", "", "specify a name")
-	nameCmd.Flags().StringVarP(&clientDo.DataFlag, "data", "d", "", "specify some data")
+	nameCmd.Flags().StringVarP(&clientDo.DataFlag, "data", "", "", "specify some data")
 	nameCmd.Flags().StringVarP(&clientDo.DataFileFlag, "data-file", "", "", "specify a file with some data")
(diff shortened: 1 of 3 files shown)


client/cmd/transaction.go
```diff
@@ -76,13 +76,13 @@ func buildTransactionCommand() {
...
 		Run:   func(cmd *cobra.Command, args []string) {
-			// transaction.Call(clientDo)
+			transaction.Call(clientDo)
 		},
 		PreRun: assertParameters,
 	}
 	callCmd.Flags().StringVarP(&clientDo.AmtFlag, "amt", "a", "", "specify an amount")
 	callCmd.Flags().StringVarP(&clientDo.ToFlag, "to", "t", "", "specify an address to send to")
-	callCmd.Flags().StringVarP(&clientDo.DataFlag, "data", "d", "", "specify some data")
+	callCmd.Flags().StringVarP(&clientDo.DataFlag, "data", "", "", "specify some data")
```
silasdavis: If you're killing the shorthand name, shouldn't you be using `StringVar`. Looking at the source I see this has the same effect (https://github.com/ogier/pflag/blob/master/string.go#L34) but you intention would be clearer and you would not be exposed to a change of implementation giving you undefined behaviour if you used `StringVar`

silasdavis:
I'm not entirely sure what this is for, so I will leave @benjaminbollen to do final signoff

client/cmd/transaction.go
benjaminbollen: P for cobra is persistent, and the need to remove d compared to old mint-client is due to introduction of -d --debug on top level command

silasdavis: OK, just saying `StringVar` is more appropriate here than `StringVarP`

benjaminbollen:
will track in #264

PR #262, opened 2016-09-07
Provide backwards support to 0.11.4 by maintaining deployment branch

benjaminbollen:
circle: add backwards support to deployment of 0.11.4 branch

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -37,3 +37,8 @@ deployment:
     commands:
       - DOCKER/build.sh
       - docker push quay.io/eris/erisdb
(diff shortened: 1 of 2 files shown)


benjaminbollen:
@pietv @compleatang does this make sense to you?  -- happy to receive feedback.  If we do something similar, then we'll best do it for e-cm (and e-pm, maybe not needed for cli?)

compleatang:
++ I think this is a good paradigm that we should be adopting for all the major libraries moving forward.

pietv:
:+1:

PR #266, opened 2016-09-09
Fix blocking event subscription in transactAndHold (master)

silasdavis:
This resolves: https://support.erisindustries.com/helpdesk/tickets/356

The issue was that we are blocking in an event callback when we try to unsubscribe, which breaks go-events contract: https://github.com/eris-ltd/eris-db/blob/master/Godeps/_workspace/src/github.com/tendermint/tendermint/events/events.go#L157 (handily found in a comment with three **X**s!)

Do we want to make this a hotfix? 

I have tested this using the code from the support ticket. It would be nice to cement this with a unit test. This could go in `test/transacting/transacting_test.go`. But it looks to me like mocking this out will be a far amount of effort and I am conscious of other priorities given that we will probably replace go-events. Open to suggestions.

Will port this to develop, and I will also make a suggestion for improvement to go-events.

### DIFF
--- a/erisdb/pipe/transactor.go
+++ b/erisdb/pipe/transactor.go
@@ -177,12 +177,25 @@ func (this *transactor) TransactAndHold(privKey, address, data []byte, gasLimit,
 	} else {
 		addr = address
 	}
-	wc := make(chan *types.EventDataCall)
+
+	// We want non-blocking on the first event received (but buffer the value),
+	// after which we want to block (and then discard the value - see below)
+	wc := make(chan *types.EventDataCall, 1)
+
 	subId := fmt.Sprintf("%X", rec.TxHash)
 	this.eventEmitter.Subscribe(subId, types.EventStringAccCall(addr), func(evt types.EventData) {
 		event := evt.(types.EventDataCall)
 		if bytes.Equal(event.TxID, rec.TxHash) {
-			wc <- &event
+			//  Beware the contract of go-events subscribe is that we must not be
+			// blocking in an event callback when we try to unsubscribe!
+			// We work around this by using a non-blocking receive.
+			select {
+			// This is a non-blocking send, but since we are using a buffered
+			// channel of size 1 we will always grab our first event even if we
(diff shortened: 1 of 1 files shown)


benjaminbollen:
FIPF, I'm not sure I fully understand what is happening under the hood of the event system that this is the sensible solution, but it does not seem to be breaking and can make sense.

silasdavis:
The core problem is that `FireEvent` assumes that a listener callback will not block indefinitely. But in our case we have following execution:

1. First matching AccCall event arrives
2. Blocking send of event data into channel `wc`
3. Receive of event data on `wc`
4. Callback unblocks
5. Second matching AccCall event arrives
6. Blocking send of event data into channel `wc`
7. No further reads on `wc`, callback remains blocking... (`FireEvent` is blocked)
8. transactAndHold tries to unsubscribe but can't because it needs the same mutex `FireEvent` has locked

So what we have is doubly bad:
1. The execution of the VM is halted waiting for the return of `FireEvent`
2. transactAndHold is blocked waiting for `Unsubscribe` to return which is deadlocked over a shared mutex with `FireEvent`

Here's my suggestion to go-events to fix it: https://github.com/tendermint/go-events/pull/6

Also just thinking about our code here.... The reason this arises is partly because we are only expecting to see one AccCall event with matching transaction receipt hash, but we see two. we ought to be able to better distinguish the original call made by the transaction and the subsequent ones. The contract that shows the bug is:

```
contract MyTest{                                          
	bool 
	
	function call1(address contractB)public{
		result = MyTest(contractB).call2(this);
	}
	
	function call2(address contractA)public returns(bool){
		bool mycall = MyTest(contractA).call3();
		return mycall;
	}
	
	function call3()public returns(bool){
		return true;
	}

	function getResult()public returns(bool){
		return result;
	}
}
```

Perhaps thinking about this is for another day, but it maybe raises the question of what are are the proper 'fire and wait' semantics. Do we just want the first call event triggered, the last?

PR #267, opened 2016-09-09
Fix blocking event subscription in transactAndHold (0.11.x)

silasdavis:
Same as: https://github.com/eris-ltd/eris-db/pull/266 for 0.11-series


benjaminbollen:
FIPF, same hotfix as applied on masterr in #266

PR #268, opened 2016-09-09
Fix blocking event subscription in transactAndHold (develop)

silasdavis:
Same as: #266 for develop


benjaminbollen:
FIPF, same hotfix as applied on master in #266

PR #269, opened 2016-09-09
Make sure command-line arguments expanded correctly

silasdavis:
Minor thing while I notice. In magic bash-land quoting the at-sign makes sure arguments containing spaces are properly expanding in the expansion when they are passed to the underlying command.


benjaminbollen:
LGTM

PR #272, opened 2016-09-12
Might be running the integration test wrapper when we didn't intend to and multiple times

silasdavis:
This might help with the bind: port already in use issues we sometimes see in circle.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -42,9 +42,9 @@ test:
   override:
     # We only wish to test our packages not vendored ones
     - echo "Running unit tests..."
-    - cd $GOPATH_REPO && glide novendor | xargs go test -v
+    - cd $GOPATH_REPO && go test -v `glide novendor`
     - echo "Running integration tests..."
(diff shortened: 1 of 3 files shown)


silasdavis:
Hopefully I was just tripping up by running the wrapper more times than the one time we need. Each time it's run is an opportunity to try and re-bind the same port where the OS may not have got round to fully un-binding...

silasdavis:
I'll re-run the circle build a few tens of times and see what happens...

silasdavis:
So first bind-related failure in a run of 9 consecutive successes then 1 failure:

```
panic: Paniced on a Crisis: listen tcp 0.0.0.0:46656: bind: address already in use
```

I still thing this has stabilised it. I am suspicious that what we are seeing might relate to something weird on circle's host/hypervisor and how it schedules freeing up ports... @benjaminbollen if you find these intermittent failures too annoying, or they get worse, then we can remove integration tests from circle. For me my (weakly held) preference would be to keep them enabled for now. If we could merge this regardless I think it improves matters.

benjaminbollen:
FIPF

PR #273, opened 2016-09-12
RPC test for BlockchainInfo

### LINKED ISSUE #218: Unit test blockchain

silasdavis:
fixes #218


silasdavis:
This also now addresses some issues in the event-based testing framework, so that it will wait for multiple blocks, and fixes some subtler issues around finality.

silasdavis:
@benjaminbollen this is now passing, if you could review

benjaminbollen:
seems this test first tests for blockheight, and as you've commented we have two different rpc functions under the same name returning a different info, so FIPF

PR #276, opened 2016-09-14
Refactor to consolidate consensus and blockchain at an appropriate level

silasdavis:
> _It hurts less if you pull it off quickly_

This refactor clears the way to implement the remaining unimplemented RPC methods. It consolidates disparate and confusing `Consensus` interfaces into a single interface and allows the `Blockchain` to initialised at the level of `Pipe` initialisation in `NewCore`.

It absorbs the previous eris-mint/blockchain struct into a simpler top level interface (a `BlockStore` with a `ChainId`) and has `Tendermint`(I dropped the 'Node') implement both `Consensus` (I dropped the 'Engine') and `Blockchain`.

As Ben originally suggested the blockchain package because a supporting actor to the consensus and manager packages, much like the account and permission packages.

Implementing the remaining RPC methods will be a matter of adding some methods to the Consensus and Blockchain interfaces and implementing them in Tendermint.


arch/constants.go
```diff
@@ -0,0 +1,9 @@
+// +build !arm
```
benjaminbollen: as discussed, maybe we can put these constants in `/util/architecture/constants.go`

blockchain/filter.go
```diff
@@ -0,0 +1,209 @@
+package blockchain
```
benjaminbollen: a general note, I took the custom of adding the GLPv3 header on every "new" file (also as a marker of what was edit during the great-reordering)

benjaminbollen:
Somewhere between FIPF, LGTM and "let's just run with it"

PR #278, opened 2016-09-15
Status() added along with a helpful gofmt

VoR0220:
Signed-off-by: VoR0220 <rj@erisindustries.com>


client/client.go
```diff
@@ -23,13 +23,16 @@ import (
...
+	rpc_types "github.com/eris-ltd/eris-db/rpc/tendermint/core/types"
 	"github.com/eris-ltd/eris-db/txs"
 )
 
 type NodeClient interface {
 	Broadcast(transaction txs.Tx) (*txs.Receipt, error)
 
 	GetAccount(address []byte) (*account.Account, error)
+
+	Status() (*rpc_types.ResultStatus, error)
```
benjaminbollen: Thanks RJ,  very nice, just one remark here: at ErisNodeClient we do not want to expose internal types (that will be under heavy change like rpc/tendermint/core/types).  So the point of these functions here is to unwrap (in this case) `rpc_types.ResultStatus`) into only fundamental types (for eris-db permissions, transactions, accounts are toplevel types; but other than that we'll prefer bytes, int etc). So I'd suggest that we even leave out NodeInfo for the time being, and re-expose that in it's own function

```Status() (ChainId []byte, ValidatorPublicKey []byte, LatestBlockHash []byte, BlockHeight uint64, LatestBlockTime int64, err error)```

benjaminbollen:
LGTM; will merge but take note that I'll change the signature of that function in a bit !

PR #280, opened 2016-09-16
Implement missing RPC methods

silasdavis:
I have tried to split this PR into three logical commits:

The primary one is the first commit implementing the structure and code paths for the unimplemented RPC methods.

The second tidies up some panics in tx encoding.

The third rationalise the layout of the tests and makes them work with the new structure.

My next PR will have tests for ListUnconfirmedTransactions, and hook in functionality to transactor in rpc/v0.


rpc/tendermint/core/routes.go
```diff
@@ -211,5 +209,38 @@ func (tmRoutes *TendermintRoutes) BlockchainInfo(minHeight,
 	} else {
 		return r, nil
 	}
+}
 
+func (tmRoutes *TendermintRoutes) ListUnconfirmedTxs() (ctypes.ErisDBResult, error) {
+	// Get all Txs for now
+	r, err := tmRoutes.tendermintPipe.ListUnconfirmedTxs(0)
```
benjaminbollen: As discussed, per incorrect comment in tendermint, `-1` is uncapped, all unconfirmed transactions; 0 means none.  But fix comes in next pr

PR #282, opened 2016-09-17
Eris client part3: QueryContract and QueryContractCode

benjaminbollen:
IMPORTANT: pulls in early integration test framework, which is incomplete but should not cause problems.  Review that it does not though.

client: correction to Status, and QueryContract and QueryContractCode


silasdavis:
LGTM

PR #286, opened 2016-09-18
Eris client part3

benjaminbollen:


rpc/tendermint/core/types/responses_util.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"github.com/eris-ltd/eris-db/core/types"
+)
+
+// UnwrapResultDumpStorage does a deep copy to remove /rpc/tendermint/core/types
+// and expose /core/types instead.  This is largely an artefact to be removed once
+// go-wire and go-rpc are deprecated.
+// This is not an efficient code, especially given Storage can be big.
+func UnwrapResultDumpStorage(result *ResultDumpStorage) (*types.Storage) {
```
benjaminbollen: 1. this needs to be tested; I do not like golang, and this is both inefficient, ugly - and probably also wrong
2. I do hope that you see the conflict I try to address here; basically we moved the two rpc's into eris-db; but I want to start opening the path to having a single (versioned) rpc - essentially deprecating /rpc/tendermint;
therefore, I dont want to expose these internal types that will disappear (sooner, rather than later)
-- the deep copy is very inefficient, but right now I want to keep moving forward and solve the problem at the core: reclaim our types properly.
-- reclaiming our types currently is tricky because we depend on go-wire, and I not enough faith in my understanding of go-wire to at this point change the encoding of the types.; Instead I want to move to gRPC+protobuf (eg) then remove this all together.

silasdavis: I'm fine with this. We are going to have to do a lot of type mappings (one-way or otherwise) so we might want to put some machinery around it. Ultimately if we are going to have our own types fed from other systems we need to create our object, particularly if they are coming from another process. So I don't see that it's so inefficient. We can replace blind mappings with more selective ones as we go.

PR #290, opened 2016-09-21
Eris client part5

benjaminbollen:


client/client.go
```diff
@@ -108,16 +93,13 @@ func (erisNodeClient *ErisNodeClient) Status() (ChainId []byte, ValidatorPublicK
...
 // QueryContract executes the contract code at address with the given data
-func (erisNodeClient *ErisNodeClient) QueryContract(address, data []byte) (ret []byte, gasUsed int64, err error) {
+// NOTE: there is no check on the caller; 
+func (erisNodeClient *ErisNodeClient) QueryContract(callerAddress, calleeAddress, data []byte) (ret []byte, gasUsed int64, err error) {
 	client := rpcclient.NewClientURI(erisNodeClient.broadcastRPC)
-	var callerAddress = make([]byte, len(address))
-	var calleeAddress = make([]byte, len(address))
-	copy(callerAddress, address)
-	copy(calleeAddress, address)
 	callResult, err := tendermint_client.Call(client, callerAddress, calleeAddress, data)
```
silasdavis: I'm must say I prefer `fromAddress` `toAddress`, maybe it's just me, but I always need to think more than I'd like when I see 'callerAddress' and 'calleeAddress'. It's like we are doing a 'Call', so we're a caller, but the caller is the contract we want to make do a call, and then callee is a sort  of an unusual term in everyday English. Subjective I guess. FIPF.

PR #295, opened 2016-09-23
client: add status command

benjaminbollen:


client/client.go
```diff
@@ -81,7 +81,7 @@ func (erisNodeClient *ErisNodeClient) Broadcast(tx txs.Tx) (*txs.Receipt, error)
 // Status returns the ChainId (GenesisHash), validator's PublicKey, latest block hash
 // the block height and the latest block time.
 func (erisNodeClient *ErisNodeClient) Status() (ChainId []byte, ValidatorPublicKey []byte, LatestBlockHash []byte, LatestBlockHeight int, LatestBlockTime int64, err error) {
-	client := rpcclient.NewClientURI(erisNodeClient.broadcastRPC)
+	client := rpcclient.NewClientJSONRPC(erisNodeClient.broadcastRPC)
```
benjaminbollen: I started out using http uri requests over go-rpc; but then I get a conflict with the prefix; currently testing at client and node side, where the rpc breaks

client/methods/status.go
```diff
@@ -0,0 +1,41 @@
...
+	log "github.com/eris-ltd/eris-logger"
+
+	"github.com/eris-ltd/eris-db/client"
+	"github.com/eris-ltd/eris-db/definitions"
+)
+
+func Status(do *definitions.ClientDo)  {
+	erisNodeClient := client.NewErisNodeClient(do.NodeAddrFlag)
+	chainId, validatorPublicKey, latestBlockHash, latestBlockHeight, latestBlockTime, err := erisNodeClient.Status()
+	if err == nil {
```
silasdavis: Not sure if we want to merge this, but this is wrong

benjaminbollen: we don't want to merge this; i put it out here to have intermediate access to the eris-client for early testing; and yes, nice catch  ;)

silasdavis:
Happy to close this?

client/methods/status.go
benjaminbollen: I've updated this, but not sure how to reflect this in the Github review; maybe a bug?

benjaminbollen:
`eris-client status` is a quick draft / hack but just to test without eris-pm 
 but that gives the correct rpc for chainId and genesishash

PR #297, opened 2016-09-23
Implemented rpc/v0 network info

### LINKED ISSUE #288: Implement .getNetworkInfo for rpc/v0
This function is used by rpc/v0 to provide network info:

https://github.com/eris-ltd/eris-db/blob/develop/manager/eris-mint/net.go#L42

It is currently a stub. It needs to be implemented. rpc/tendermint has a similar (but implemented):

https://github.com/eris-ltd/eris-db/blob/develop/manager/eris-mint/pipe.go#L310

Only this does not include 'moniker'. It probably should.

silasdavis:
Didn't realise this was missing until recently. Easy enough to implement for release.

fixes #288


definitions/pipe.go
```diff
@@ -68,16 +67,6 @@ type NameReg interface {
 	Entries([]*event.FilterData) (*types.ResultListNames, error)
 }
 
-type Net interface {
```
silasdavis: not used outside of v0, no need to have this on Pipe since underlying information replicated with Peers, Listeners etc. make way for RPC unification

silasdavis:
I've deduplicated the `Peer` struct type, and made other objects reference it via a pointer as well as methods return a pointer value, as is more in keeping with other types.

Also fixed mocks.

benjaminbollen:
Not to keen on `/rpc/v0/shared`, but very good work on starting to integrate the two rpcs which we definitely want to push forward! so "looks very good to me"!  

We need integration tests to fully sign off, but the updated unit tests and local testing is good for now

silasdavis:
Yeah agreed on /rpc/v0/shared. But I see it as a very temporary stopgap pending the sorting out of RPC. I had a 'pipes' before that was essentially serving as a v0 shared namespace (for restServer.go and methods.go) just to avoid duplication.

PR #304, opened 2016-09-29
Eris node websocket client

benjaminbollen:


client/websocket_client.go
```diff
@@ -0,0 +1,136 @@
...
+				continue
+			}
+
+			data, ok := result.Data.(txs.EventDataTx)
+			if !ok {
+				// We are on the lookout for EventDataTx
+				confirmationChannel <- Confirmation{
+					BlockHash: latestBlockHash,
+					Event: nil // or result.Data ?
+					Exception: fmt.Errorf("response error: expected result.Data to be *types.EventDataTx")
```
benjaminbollen: to be continued here

client/websocket_client.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+
+	}()
+
+	// TODO: [ben] this is a draft implementation as resources on time.After can not be
+	// recovered before the timeout.  Close-down timeout at success properly.
+	timeout := time.After(time.Duration(MaxCommitWaitTimeSeconds) * time.Second)
+
+	go func() {
+		<- timeout
```
silasdavis: I wonder if we should use something like:

```golang
go func(){
    select {
    case <-time.After(time.Duration(MaxCommitWaitTimeSeconds) * time.Second):
        // Error confirmation
        return
    case <-shutdownCh:
        // Avoid timing out after have received confirmation
    }
}
```

Where shutdownCh gets sent a message as soon as we have a valid confirmation.

As it stands It shouldn't really make any difference, just if code were to be added before the return of the function there could be a race.

Hmm... I'm not sure if this is a valid thing to do really in the circumstances, but I've suggested it now so I'll leave it!

keys/key_client.go
```diff
@@ -17,13 +17,14 @@
...
 import (
+	"fmt"
 	"encoding/hex"
 )
 
 type KeyClient interface {
 	// Sign needs to return the signature bytes for given message to sign
 	// and the address to sign it with.
-	Sign(signBytes []byte, signAddress []byte) (signature []byte, err error)
+	Sign(signBytesString string, signAddress []byte) (signature []byte, err error)
```
silasdavis: I take it this is just because the way things are... If so, noted and I move on

client/websocket_client.go
```diff
@@ -0,0 +1,193 @@
...
+// Returns a channel that will receive a confirmation with a result or the exception that
+// has been confirmed; or an error is returned and the confirmation channel is nil.
+func (erisNodeWebsocketClient *ErisNodeWebsocketClient) WaitForConfirmation(tx txs.Tx, chainId string, inputAddr []byte) (chan Confirmation, error) {
+	// check no errors are reported on the websocket 
+	if err := erisNodeWebsocketClient.assertNoErrors(); err != nil {
+		return nil, err
+	}
+
+	// Setup the confirmation channel to be returned
+	confirmationChannel := make(chan Confirmation, 1)
```
benjaminbollen: strictly speaking it is possible to get multiple confirmation on this channel; so strictly speaking the receiver needs to empty it.  It's very much a draft construction but close/if not literal from mint-client.  A lot of room for improvement

client/websocket_client.go
```diff
@@ -0,0 +1,193 @@
...
+			if wire.ReadJSONPtr(result, resultBytes, &err); err != nil {
+				// keep calm and carry on
+				log.Errorf("eris-client - Failed to unmarshal json bytes for websocket event: %s", err)
+				continue
+			}
+			
+			event, ok := (*result).(*ctypes.ResultEvent)
+			if !ok {
+				// keep calm and carry on
+				log.Error("eris-client - Failed to cast to ResultEvent for websocket event")
```
benjaminbollen: yes, but it should only be `EventDataTx` in the case of the `InputAddr` as event string

benjaminbollen: so you would log warning?

benjaminbollen: You're right that this indeed is not an error;

client/websocket_client.go
```diff
@@ -0,0 +1,193 @@
...
+				continue
+			}
+
+			if event.Event != eid {
+				log.Warnf("Received unsolicited event! Got %s, expected %s\n", event.Event, eid)
+				continue
+			}
+
+			data, ok := event.Data.(txs.EventDataTx)
+			if !ok {
```
benjaminbollen: Here I think it *should* only be an EventDataTx for the event string formed from `InputAddr`; so this is stronger an error case than above I think

client/websocket_client.go
```diff
@@ -0,0 +1,193 @@
...
+				Error: nil,
+			}
+			return
+		}
+
+	}()
+
+	// TODO: [ben] this is a draft implementation as resources on time.After can not be
+	// recovered before the timeout.  Close-down timeout at success properly.
+	timeout := time.After(time.Duration(MaxCommitWaitTimeSeconds) * time.Second)
```
benjaminbollen: Ha! To my victory.  The mint-client code had time.Tick which I thought was more concerning: if used as a command line, then that is fine as program terminates; but we use it as a library in e-pm and that is a problematic use (of time.Tick here);  So I changed it to time.After, so indeed that should saturate the resources used on a 10 second window; but will still pile up these go-routines if many calls are made per second.

For now this can pass as sufficient, but for load testing, I think this might be a point of concern

client/websocket_client.go
benjaminbollen: It s a good suggestion; there are more edge-cases so probably this whole architecture can be drafted better

keys/key_client.go
benjaminbollen: the way things are :)

client/websocket_client.go
silasdavis: Do you want to remove it or downgrade to Info?

client/websocket_client.go
silasdavis: You still might as well in-line this..

PR #305, opened 2016-09-29
Don't use difference between caller's and callee's gas as the gas limit for the call

silasdavis:
UPDATE] (04/10/2016) - seems this is a legitimate fix for https://github.com/eris-ltd/eris-pm/issues/212, so:
fixes https://github.com/eris-ltd/eris-pm/issues/212

I had some trouble convincing myself this could really be a bug, because as far as I can see it will have been a bug pretty much forever, and it's kind of fundamental. I think the reason it was not spotted is because we generally make very generous gas allowances, and for reasons I don't understand when EVM `CALL`ing (and friends) the gas allowance for those calls (allowed to the callee) must have been small to that available for the caller.

Explanation:
At the beginning of executing `CALL` and friends (`CALLCODE`, `DELEGATECALL`) we pop off a gasLimit value that the callee should be allotted for its execution:

```go
gasLimit := stack.Pop64()
```

We make sure the callee is not asking for more gas than the caller, and provided it's not, we pre-emptively remove the full allowance before any computation has occurred (don't worry we'll pay it back later).
```go
			// Ensure that gasLimit is reasonable
			if *gas < gasLimit {
				return nil, firstErr(err, ErrInsufficientGas)
			} else {
				*gas -= gasLimit
				// NOTE: we will return any used gas later.
			}
```

But then when we make the actual child call we use `gas`, which is now the _difference_ between what caller has allotted and what the callee should be allowed:

```go
ret, err = vm.Call(callee, acc, acc.Code, args, value, gas)
```

Suppose the caller has 10000 gas, and the caller is allotted (a `gasLimit` of) 9999, then rather than calling with 9999 gas, we are calling with just 1!

For our native contracts we pass a pointer to `gasLimit` (as we should do):

```go
ret, err = nativeContract(vm.appState, callee, args, &gasLimit)
```

And later we add on the remainder (that wasn't used by the callee):

```go
			// Handle remaining gas.
			*gas += gasLimit
```

Which seems suggestive enough that this has indeed always been wrong. Unless I'm missing something, which I might be. But I think we should be passing the `gasLimit` we popped off the stack.

### DIFF
--- a/Godeps/_workspace/src/github.com/tendermint/tendermint/vm/test/vm_test.go
+++ b/Godeps/_workspace/src/github.com/tendermint/tendermint/vm/test/vm_test.go
@@ -8,12 +8,19 @@ import (
 	"testing"
 	"time"
 
+	"errors"
+
 	. "github.com/eris-ltd/eris-db/Godeps/_workspace/src/github.com/tendermint/tendermint/common"
 	"github.com/eris-ltd/eris-db/Godeps/_workspace/src/github.com/tendermint/tendermint/events"
 	ptypes "github.com/eris-ltd/eris-db/Godeps/_workspace/src/github.com/tendermint/tendermint/permission/types"
 	"github.com/eris-ltd/eris-db/Godeps/_workspace/src/github.com/tendermint/tendermint/types"
 	. "github.com/eris-ltd/eris-db/Godeps/_workspace/src/github.com/tendermint/tendermint/vm"
+	"github.com/stretchr/testify/assert"
 )
+
+func init() {
+	SetDebug(true)
+}
 
 func newAppState() *FakeAppState {
 	fas := &FakeAppState{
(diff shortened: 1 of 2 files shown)


silasdavis:
To add to why this might not have been spotted before, and was for Jan's issue: https://github.com/eris-ltd/eris-pm/issues/212. Is that using a `mapping` triggers some EVM code calling a SHA3 hash which incurs a gas cost related to its length. Whereas most of our gas prices are set very low (lower than Ethereum).

silasdavis:
Don't merge yet. If it looks like the right thing I will be adding a unit_test to cover the issue in vm_test.go. Will also need to port to all other versions.

benjaminbollen:
This is very interesting.  I can remark that while debugging the evm before I did notice an unusual jump in the gas, but did not investigate further; only strikes as suspicious now.  I will have to write out a simple example for myself later.  Very nice analysis

benjaminbollen:
Very nice work @silasdavis. After following your argument and the code myself, I am equally confident that this is - regardless of whether it fixes https://github.com/eris-ltd/eris-pm/issues/212 - a legitimate bug that was there.  The expanded use of library contracts has started to showcase the bug which was unimportant because indeed we are generous with the initial available gas.

(another indicator is that it was correctly implemented for native contracts (on https://github.com/eris-ltd/eris-db/pull/305/files#diff-cad6213f6a08c87e4110c4006789e2b7R811).  

If you want to add an evm unit test (or an rpc integration test, if the event structure can't easily be mocked out), then that is great too.

LGTM. ready to merge and pull to master -rc2 and -rc3 when you're happy with it.

silasdavis:
I was going to add a test in vm_test.go I just want to get an idea of what real solc-generated byte code looks like first

silasdavis:
I've added in a test that would have caught the old bug, and also exercises DELEGATECALL a bit

benjaminbollen:
LGTM, the bug fix is very well argued and consistently observed once pointed out by Silas.  The test is more complex to fully verify, but is well-constructed and passes, so thumbs up from me.

PR #306, opened 2016-10-04
Use gasLimit for DELEGATECALL and test

silasdavis:
As: https://github.com/eris-ltd/eris-db/pull/305


silasdavis:
My changes cause a fault in state_test.go

PR #313, opened 2016-10-06
Let's not depend on docker

silasdavis:
Goddamn auto-import in Intellij loves to choose docker packages over stdlib, so that's why glide thought we had a docker dependency.

This madness must stop.


benjaminbollen:
LGTM!

PR #314, opened 2016-10-06
We were (probably) inadvertently depending on Tendermint's native RPC

silasdavis:
I noticed that transactor was relying on a Tendermint native RPC call `broadcast_tx_sync`:

```
_, err = app.client.Call("broadcast_tx_sync", params, &result)
```

One that we no longer support. We could implement it, or we could port it to call our RPC, but then this would still be one RPC making another HTTP request to talk to another RPC in the same process, which makes little sense. It ended up this way as we adopted a cautious approach in the refactor.

I have elected to instead delegate to our existing tested BroadcastTx implementation.

...

This also lead me to look at the history of the 'tendermint' endpoint (running on 46657 by default). This RPC was split into our 'rpc/tendermint' and Tendermint's own native RPC. Since we are running in-process we have no real need to run Tendermint's RPC (and it seems to me less can go wrong if we don't), but transactor was still expecting a Tendermint endpoint (though it looked similar to ours).

I then noticed that we are actually trying to run both our rpc/tendermint and Tendermint's native RPC on the same address and port (46657 by default). This obviously will not work. I think the best thing is to not run the Tendermint native one at all for now.

I have tried to take an approach that either simplifies or leaves things alone. I was tempted to rename/rearrange some config, but the fix works without that so I have avoided it.

### DIFF
--- a/consensus/tendermint/config.go
+++ b/consensus/tendermint/config.go
@@ -73,7 +73,7 @@ func (tmintConfig *TendermintConfig) AssertTendermintDefaults(chainId, workDir,
 	tmintConfig.SetDefault("db_backend", "leveldb")
 	tmintConfig.SetDefault("db_dir", dataDir)
 	tmintConfig.SetDefault("log_level", "info")
-	tmintConfig.SetDefault("rpc_laddr", "0.0.0.0:46657")
+	tmintConfig.SetDefault("rpc_laddr", "")
 	tmintConfig.SetDefault("prof_laddr", "")
 	tmintConfig.SetDefault("revision_file", path.Join(workDir, "revision"))
 	tmintConfig.SetDefault("cswal", path.Join(dataDir, "cswal"))
@@ -103,6 +103,7 @@ func (tmintConfig *TendermintConfig) AssertTendermintConsistency(
 	tmintConfig.Set("genesis_file", consensusConfig.GenesisFile)
 	// private validator file
 	tmintConfig.Set("priv_validator_file", privateValidatorFilePath)
+
 }
 
 // implement interface github.com/tendermint/go-config/config.Config
--- a/consensus/tendermint/tendermint.go
+++ b/consensus/tendermint/tendermint.go
@@ -111,6 +111,16 @@ func NewTendermint(moduleConfig *config.ModuleConfig,
 		path.Join(moduleConfig.RootDir,
 			moduleConfig.Config.GetString("private_validator_file")))
 
+	// TODO: [Silas] we want to something better than this like not not have it in
+	// the config at all, but for now I think it's much safer to make sure we are
+	// not running the tendermint RPC as it could lead to unexpected behaviour,
+	// not least if we accidentally try to run it on the same address as our own
+	if tmintConfig.GetString("rpc_laddr") != "" {
+		log.Warnf("Force disabling Tendermint's native RPC, which had been set to " +
+				"run on '%s' in the Tendermint config.", tmintConfig.GetString("rpc_laddr"))
+		tmintConfig.Set("rpc_laddr", "")
+	}
+	
 	newNode := node.NewNode(tmintConfig, privateValidator, func(_, _ string,
 		hash []byte) proxy.AppConn {
 		return NewLocalClient(new(sync.Mutex), application)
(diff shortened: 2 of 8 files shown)


consensus/tendermint/config.go
```diff
@@ -73,7 +73,7 @@ func (tmintConfig *TendermintConfig) AssertTendermintDefaults(chainId, workDir,
 	tmintConfig.SetDefault("db_backend", "leveldb")
 	tmintConfig.SetDefault("db_dir", dataDir)
 	tmintConfig.SetDefault("log_level", "info")
-	tmintConfig.SetDefault("rpc_laddr", "0.0.0.0:46657")
```
benjaminbollen: These defaults I've taken from Tendermint; it is important to check that Tendermint will not set them, if they are not set by us.  but yes, we definitely want to look at disabling tendermints internal rpc for in-process tendermint

consensus/tendermint/tendermint.go
```diff
@@ -111,6 +111,16 @@ func NewTendermint(moduleConfig *config.ModuleConfig,
 		path.Join(moduleConfig.RootDir,
 			moduleConfig.Config.GetString("private_validator_file")))
 
+	// TODO: [Silas] we want to something better than this like not not have it in
+	// the config at all, but for now I think it's much safer to make sure we are
+	// not running the tendermint RPC as it could lead to unexpected behaviour,
+	// not least if we accidentally try to run it on the same address as our own
+	if tmintConfig.GetString("rpc_laddr") != "" {
```
benjaminbollen: brilliant, but Im not sure that it would work;  I would almost default to setting the tendermint rpc to a different port by default `0.0.0.0:48857` eg

manager/eris-mint/pipe.go
```diff
@@ -87,34 +87,47 @@ func NewErisMintPipe(moduleConfig *config.ModuleConfig,
...
+	// NOTE: [Silas]
+	// This is something of a loopback, but seems like a nicer option than
+	// transactor calling the Tendermint native RPC (as it was before),
+	// or indeed calling this RPC over the wire given that we have direct access.
+	//
+	// We could just hand transactor a copy of Pipe, but doing it this way seems
+	// like a reasonably minimal and flexible way of providing transactor with the
+	// broadcast function it needs, without making it explicitly
+	// aware of/depend on Pipe.
+	transactor := newTransactor(moduleConfig.ChainId, eventSwitch, erisMint,
```
benjaminbollen: alternatively we can hand transactor a pointer to ConsensusEngine, when it is set; but we'll come back to this soon, and then remove the need for these compatibility constructions

server_config.toml
```diff
@@ -150,7 +150,7 @@ listener = "tcp://0.0.0.0:46658"
 # This file needs to be in the root directory
 private_validator_file = "priv_validator.json"
 
-  # Tendermint requires additional configuration parameters.
+  # Tendermint requires additional configuration parameters.TransactAndHold
```
benjaminbollen: This is a typo, but also we should no longer have server_config.toml in the root directory.

consensus/tendermint/tendermint.go
silasdavis: No it disables it: https://github.com/tendermint/tendermint/blob/master/node/node.go#L362

manager/eris-mint/pipe.go
silasdavis: Unless there is a strong reason to change it, I'd rather stick with this, as soon it will all be laid to waste as you say...

server_config.toml
silasdavis: oops yeah, I'll fix that. I more or less agree, at least not in this form. Having send that i think there is an argument to have the config templating as a part of eris-db/eris-client, so changes to config happen in the same place as they are consumed. i.e. move some of what is in eris-cm. But let's shelf this.

benjaminbollen: yeah, while I was typing that, it makes sense to me as well that having cm define the config and edb consume it is dangerous.  Better for cm to have it write and manage all, but depend on a eris-client functionality

manager/eris-mint/pipe.go
benjaminbollen: no stick with it here.  That was what I was hoping to convey.  It s very nice and like it.

benjaminbollen:
LGTM;

silasdavis:
will just make those changes

benjaminbollen:
LGTM

PR #315, opened 2016-10-06
Reintroduce Send and SendAndHold

### LINKED ISSUE #155: Upgrade existing chains to single configuration file and configure for single data directory
1. Eris-DB v0.11.4 and prior read `config.toml` by Tendermint and `server_conf.toml` by Eris-DB; as of 0.12.0-rc2 Eris-DB reads a single `server_config.toml`
2. As of Eris-DB v0.12.0rc2 the state database is (by default) separated from the consensus engine block store; but on existing chains (Eris-DB v0.11.4) these files are in the same directory; make sure Eris-DB v0.12.0 configures itself for a single shared directory for existing chains.

### LINKED ISSUE #298: erisdb.send RPC missing in rc3
```shell
curl -H "Content-Type: application/json" -X POST -d '{"jsonrpc":"2.0","method":"erisdb.send","params":{"priv_key":"3D3305845A2CDEA61CD8678E7E1945F9A56FBA8CB2DA1F5A5D06448ADDEDD5645806A8BC1372EBCA5B5D2B65B7F86298F41ABCC7E6B6CF977E25127F58268CE2","to_address":"0000000000000000000000000000000000000010","amount":42},"id":"1"}' http://192.168.99.100:32902/rpc
```

yields

```JSON
{"result":null,"error":{"code":-32601,"message":"Method not found: erisdb.send"},"id":"1","jsonrpc":"2.0"}
```

silasdavis:
Probably shouldn't have been removed.

fixes #298 
fixes #155

Note that the old implementation: https://github.com/eris-ltd/eris-db/pull/114/commits/b2d0b4d402319dbfef914354750bb69dae728025

used:
```
cache := this.mempoolReactor.Mempool.GetCache()
￼acc := cache.GetAccount(pa.Address)
```

whereas we use:

```
cache := this.erisMint.GetState()
￼acc := cache.GetAccount(pa.Address)
```

This follows Call

### DIFF
--- a/rpc/v0/params.go
+++ b/rpc/v0/params.go
@@ -90,6 +90,13 @@ type (
 		GasLimit int64  `json:"gas_limit"`
 	}
 
+	// Used when sending a 'Send' transaction.
+	SendParam struct {
+		PrivKey   []byte `json:"priv_key"`
+		ToAddress []byte `json:"to_address"`
+		Amount    int64  `json:"amount"`
+	}
+
 	NameRegEntryParam struct {
 		Name string `json:"name"`
 	}
--- a/definitions/pipe.go
+++ b/definitions/pipe.go
@@ -77,6 +77,8 @@ type Transactor interface {
 		fee int64) (*txs.Receipt, error)
 	TransactAndHold(privKey, address, data []byte, gasLimit,
 		fee int64) (*txs.EventDataCall, error)
+  Send(privKey, toAddress []byte, amount int64) (*txs.Receipt, error)
+	SendAndHold(privKey, toAddress []byte, amount int64) (*txs.Receipt, error)
 	TransactNameReg(privKey []byte, name, data string, amount,
 		fee int64) (*txs.Receipt, error)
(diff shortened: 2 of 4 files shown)


benjaminbollen:
For sake of connecting open ends; this relates to https://github.com/eris-ltd/eris-db/issues/115

silasdavis:
Ah! Well you tried!

rpc/v0/params.go
```diff
@@ -90,6 +90,13 @@ type (
 		GasLimit int64  `json:"gas_limit"`
 	}
 
+	// Used when sending a 'Send' transaction.
+	SendParam struct {
+		PrivKey   []byte `json:"priv_key"`
```
benjaminbollen: why is a private key included here?

benjaminbollen: ah never mind, transact and transactAndHold have the same "broken" architecture; both of these need to be deprectaed in favor of using eris-client which now has the `waitForConfirmation`; but we can merge it for rc3

PR #319, opened 2016-10-10
fix circle problems; push correct image

compleatang:
circle can't rmi images and so build_tool will fail.


benjaminbollen:
LGTM

PR #321, opened 2016-10-11
Feature smaller docker redux

compleatang:


benjaminbollen:
LGTM; nice catch, yes we don't want to push the build images either

PR #325, opened 2016-10-11
Bump version, but label as development version

silasdavis:
Suggest that we merge this before the 0.12.0-RC3 PR to avoid accidentally releasing over the top of the quay image in future. Also suggest that we adopt the convention of sufficing our version on the develop branch with `-develop` so that we never push to a quay tag that ought to be reserved for a real non-development version.


benjaminbollen:
I like that!

benjaminbollen:
LGTM

PR #368, opened 2016-11-07
b. issue366 only deploy from branch `release-0.12` for MASTER

### LINKED ISSUE #366: [CD] deploy from release branches
maintain `release-0.11` and `release-0.12` branches, and only deploy from pushes to these branches; decoupling the deployment from full integrated testing on master branches

benjaminbollen:
fixes #366


silasdavis:
So if master is no longer our release branch, what is for, will it hold our current most recent release branch, but release-<major>.<minor> will hold our latest patch versions for each release and be used to release?

benjaminbollen:
so the proposal is to offset the deployment cycle by one step from the development cycle; that way all repos can be on master; and that is then pushed out and ready for full network/soak testing, and after the completion of that, moving it to release, makes it the official release

silasdavis:
When you say 'deployment cycle', do you mean deployment by Jenkins to testing environments, sounds like it, but that term is a little overloaded?

benjaminbollen:
No, with deployment cycle I mean "making the official release that is used by default by our users".  It is true that we do not have the full deployment cycle, and perhaps I should call it "release cycle" instead.

Issue #371, opened 2016-11-10
[rpc v0] TransactAndHold result is affected by library calls


benjaminbollen:
😺

benjaminbollen:
hotfixed and pushed on `quay.io/eris/db:0.12.1`

benjaminbollen:
confirmed working on 0.12.1 and better fix on 0.16.0

PR #372, opened 2016-11-11
issue371: hotfix 0.12.1

benjaminbollen:
eris-mint/vm: issue371, hotfix silence event call for delegate call as no transaction can commence with delegatecall

is a hotfix for #371; but needs a more elegant solution:
```introduce event EventStringAccDelegateCall Acc/%X/DelegateCall```


silasdavis:
It's dead Jim. Yeah should have own event. LGTM

PR #412, opened 2016-11-29
Issue 411 structure build and test pipeline

benjaminbollen:
- introduce makefile for eris-db with 
`make check` : fails on gofmt
`make fmt` : applies gofmt
`make lint` : fails on golint
`make vet` : fails on go vet
`make erase_vendor` : deletes `/vendor`, as helper to ensure CI builds have clean vendor folder installed
`make install_vendor` : installs `/vendor` from `glide.lock`
`make build` : fail on check, build all targets
`make build_race` : fail on check, build all targets with check for race conditions
`make test` : fail on check, build and run go test (for all targets)
`make test_race` : as test, but with race conditions
`make build_docker_db` : build docker image for eris-db and tag with full version number
`make test_docker_db` : build docker image for eris-db and run go test inside the docker image

note: build also builds into `/target`

this set of instructions is not set in stone or incomplete, but it aims at providing a base-line to simplify the jenkins jobs needed to build, test, move artefacts, for multiple targets in a single repository as we consolidate the repositories.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,9 @@
 # do not vendor in dependencies
 vendor
+
+# build target
+target
+
 # Temporary / cached
 *.swp
 debug
--- a/Dockerfile.deploy
+++ b/Dockerfile.deploy
@@ -3,13 +3,8 @@ MAINTAINER Monax <support@monax.io>
 
 ENV TARGET eris-db
 
-# runtime customization start here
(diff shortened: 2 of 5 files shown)


benjaminbollen:
@silasdavis I started using a feature branch to allow for more flexibility for discussion, rather than requesting a merge into develop straight away.

benjaminbollen:
:smiley_cat:

PR #416, opened 2016-12-06
Pull in configuration templating from eris-cm

silasdavis:
When eris-cm is deprecated cli chould depend on eris-db to generate its config, so that eris-db becomes the source of truth on its own config.


zramsay:
this should also be added to `feature_consolidate` branch

benjaminbollen:
LGTM

benjaminbollen:
@zramsay yes, but it is a disconnected package at this point, so we can merge develop into consolidate branch

PR #428, opened 2016-12-20
Logging conventions etc

silasdavis:


logging/metadata.go
```diff
@@ -5,6 +5,7 @@ import (
 
 	"github.com/eris-ltd/eris-db/logging/loggers"
 	"github.com/eris-ltd/eris-db/logging/structure"
+	"github.com/eris-ltd/mint-client/Godeps/_workspace/src/github.com/inconshreveable/log15/stack"
```
silasdavis: well spotted, intellij auto-import, ugh

silasdavis:
@benjaminbollen fixed this. Apparently slices aren't hashable so die if you try to use them as a key for a map.

PR #442, opened 2017-01-13
Add documentation comment on lazily initialisation of account public keys

silasdavis:


manager/eris-mint/state/execution.go
```diff
@@ -203,6 +203,12 @@ func getOrMakeOutputs(state AccountGetter, accounts map[string]*acm.Account, out
 	return accounts, nil
 }
 
+// Since all ethereum accounts implicitly exist we sometimes lazily create an Account object to represent them
+// only when needed. Sometimes we need to create an unknown Account knowing only its address (which is expected to
+// be a deterministic hash of its associated public key) and not its public key. When we eventually receive a
+// transaction acting on behalf of that account we will be given a public key that we can check matches the address.
+// If it does then we will associate the public key with the stub account already registered in the system once and
+// for all time.
```
benjaminbollen: Note: worth adding that for code contracts the address of an account is derived not from a public key - it is derived by the evm or even (arbitrarily) provided in a CALL transaction to this new account, and it is unfeasible to search for a private-public key pair that can unlock the account.  Such an account will only be controlled by the code it has stored.

silasdavis: Hm, it is arbitrarily _provided_, which means the actor triggering the call has the option of deriving it from a public key beforehand. I am not aware of a place where the EVM 'generates' it... If it does can you link?

benjaminbollen: on EVM CREATE or if a CallTx goes to an uninitialised Account (state)

PR #447, opened 2017-01-17
Remove eris-logger references from eris-db server

silasdavis:
Little incremental refactor stripping out eris-logger

### DIFF
--- a/event/event_cache.go
+++ b/event/event_cache.go
@@ -70,7 +70,7 @@ func reap(es *EventSubscriptions) {
 	defer es.mtx.Unlock()
 	for id, sub := range es.subs {
 		if time.Since(sub.ts) > reaperThreshold {
-			// Seems like Go is ok with this..
+			// Seems like Go is ok with evts..
 			delete(es.subs, id)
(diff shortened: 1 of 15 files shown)


silasdavis:
@VoR0220 this change entails no tests. It's just changing existing logging to use new framework. There are some tests in the logging framework itself which is already merged to this feature branch.

VoR0220:
@silasdavis what I mean is, are all the chains server tests cleaned of the eris-logger?

silasdavis:
Can someone press the button

event/event_cache.go
```diff
@@ -70,7 +70,7 @@ func reap(es *EventSubscriptions) {
 	defer es.mtx.Unlock()
 	for id, sub := range es.subs {
 		if time.Since(sub.ts) > reaperThreshold {
-			// Seems like Go is ok with this..
+			// Seems like Go is ok with evts..
```
benjaminbollen: You were a bit quick with the replace alls for `this` -> `evts`

silasdavis: fixed in current working directory

PR #450, opened 2017-01-18
Remove eris-logger from client

silasdavis:
This threads the logging through eris-client.

Mostly the same functionality (although there is a lack of tests in this repo so will need testing with eris-pm), but some slight behavioural changes possible where I have opted to exit fatally or removed some log output for failures of basic initialisation.

I've made lacking tcp:// or unix:// for node address fatal, but I think it should maybe just normalise/default to tcp://. Also we might want to parse it as a URL.

### DIFF
--- a/client/cmd/transaction.go
+++ b/client/cmd/transaction.go
@@ -17,14 +17,12 @@
 package commands
 
 import (
-	"os"
-	"strings"
+"strings"
 
 	"github.com/spf13/cobra"
 
-	log "github.com/eris-ltd/eris-logger"
-
 	"github.com/eris-ltd/eris-db/client/methods"
+	"github.com/eris-ltd/eris-db/util"
 )
 
 func buildTransactionCommand() *cobra.Command {
@@ -180,26 +178,21 @@ func defaultAddress() string {
 
 func assertParameters(cmd *cobra.Command, args []string) {
 	if clientDo.ChainidFlag == "" {
-		log.Fatal(`Please provide a chain id either through the flag --chain-id or environment variable $CHAIN_ID.`)
-		os.Exit(1)
+		util.Fatalf(`Please provide a chain id either through the flag --chain-id or environment variable $CHAIN_ID.`)
 	}
 
 	if !strings.HasPrefix(clientDo.NodeAddrFlag, "tcp://") &&
-		!strings.HasPrefix(clientDo.NodeAddrFlag, "unix://") {
+			!strings.HasPrefix(clientDo.NodeAddrFlag, "unix://") {
 		// TODO: [ben] go-rpc will deprecate reformatting; also it is bad practice to auto-correct for this;
-		log.Warn(`Please use fully formed listening address for the node, including the tcp:// or unix:// prefix.`)
+		// TODO: [Silas] I've made this fatal, but I'm inclined to define the default as tcp:// and normalise as with http
+		// below
+		util.Fatalf(`Please use fully formed listening address for the node, including the tcp:// or unix:// prefix.`)
 	}
 
 	if !strings.HasPrefix(clientDo.SignAddrFlag, "http://") {
 		// NOTE: [ben] we preserve the auto-correction here as it is a simple http request-response to the key server.
+		// TODO: [Silas] we don't have logging here to log that we've done this. I'm inclined to either urls without a scheme
+		// and be quiet about it, or to make non-compliance fatal
 		clientDo.SignAddrFlag = "http://" + clientDo.SignAddrFlag
-		log.WithFields(log.Fields{
-			"signing address": clientDo.SignAddrFlag,
-		}).Warn(`Please use fully formed listening address for the key server; adding http:// prefix`)
 	}
-	log.WithFields(log.Fields{
-		"signing address": clientDo.SignAddrFlag,
-		"node address":    clientDo.NodeAddrFlag,
-		"chain id":        clientDo.ChainidFlag,
-	}).Debug("Asserted parameters")
 }
--- a/client/cmd/transaction.go
+++ b/client/cmd/transaction.go
@@ -17,7 +17,7 @@
 package commands
 
 import (
-"strings"
+	"strings"
 
 	"github.com/spf13/cobra"
 
@@ -31,9 +31,8 @@ func buildTransactionCommand() *cobra.Command {
 	transactionCmd := &cobra.Command{
 		Use:   "tx",
 		Short: "eris-client tx formulates and signs a transaction to a chain",
-		Long: `eris-client tx formulates and signs a transaction to a chain.
-`,
-		Run: func(cmd *cobra.Command, args []string) { cmd.Help() },
+		Long:  "eris-client tx formulates and signs a transaction to a chain.",
+		Run:   func(cmd *cobra.Command, args []string) { cmd.Help() },
 	}
 
 	addTransactionPersistentFlags(transactionCmd)
@@ -44,7 +43,10 @@ func buildTransactionCommand() *cobra.Command {
 		Short: "eris-client tx send --amt <amt> --to <addr>",
 		Long:  "eris-client tx send --amt <amt> --to <addr>",
 		Run: func(cmd *cobra.Command, args []string) {
-			methods.Send(clientDo)
+			err := methods.Send(clientDo)
+			if err != nil {
+				util.Fatalf("Could not complete send: %s", err)
+			}
 		},
 		PreRun: assertParameters,
 	}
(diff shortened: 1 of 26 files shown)


client/cmd/transaction.go
```diff
@@ -17,14 +17,12 @@
 package commands
 
 import (
-	"os"
-	"strings"
+"strings"
```
benjaminbollen: gofmt?

benjaminbollen:
@silasdavis as a general point you use the `util.Fatalf()` in a client library; and that should not happen; no package that is intended to be imported as a library package can have a call to `os.Exit(1)` https://github.com/eris-ltd/eris-db/blob/6c30c198b4a4e20e28c3f056dafcd7ccbc0daa76/util/os.go#L9

If you agree, can you go in and make sure that only errors are returned to the top level functions exposed?

client/cmd/transaction.go
```diff
@@ -180,26 +178,21 @@ func defaultAddress() string {
...
 	}
 
 	if !strings.HasPrefix(clientDo.NodeAddrFlag, "tcp://") &&
-		!strings.HasPrefix(clientDo.NodeAddrFlag, "unix://") {
+			!strings.HasPrefix(clientDo.NodeAddrFlag, "unix://") {
 		// TODO: [ben] go-rpc will deprecate reformatting; also it is bad practice to auto-correct for this;
-		log.Warn(`Please use fully formed listening address for the node, including the tcp:// or unix:// prefix.`)
+		// TODO: [Silas] I've made this fatal, but I'm inclined to define the default as tcp:// and normalise as with http
+		// below
+		util.Fatalf(`Please use fully formed listening address for the node, including the tcp:// or unix:// prefix.`)
```
benjaminbollen: instance of use of util.Fatalf

client/cmd/transaction.go
```diff
@@ -180,26 +178,21 @@ func defaultAddress() string {
 
 func assertParameters(cmd *cobra.Command, args []string) {
 	if clientDo.ChainidFlag == "" {
-		log.Fatal(`Please provide a chain id either through the flag --chain-id or environment variable $CHAIN_ID.`)
-		os.Exit(1)
+		util.Fatalf(`Please provide a chain id either through the flag --chain-id or environment variable $CHAIN_ID.`)
```
benjaminbollen: woops, arguably here we used os.Exit(1) -- guilty as charged

client/methods/call.go
```diff
@@ -17,31 +17,33 @@
...
 )
 
 func Call(do *definitions.ClientDo) {
 	// construct two clients to call out to keys server and
 	// blockchain node.
-	erisKeyClient := keys.NewErisKeyClient(do.SignAddrFlag)
-	erisNodeClient := client.NewErisNodeClient(do.NodeAddrFlag)
+	logger, err := loggerFromClientDo(do, "Call")
+	if err != nil {
+		util.Fatalf("Could not generate logging config from ClientDo: %s", err)
```
benjaminbollen: but the new introductions of it are, I believe, to be avoided.  I would not want to regress on this point of no forced exits from library package.

zramsay: we use `util.IfExit()` for the record. Might be good to standardize across repos.

zramsay: see https://github.com/eris-ltd/eris-cli/blob/master/util/exit.go

event/event_cache_test.go
```diff
@@ -178,7 +178,7 @@ func TestSubManualClose(t *testing.T) {
 // Test that the system doesn't fail under high pressure.
 func TestSubFlooding(t *testing.T) {
 	NUM_SUBS := 100
-	// Keep the reaper out of evts.
+	// Keep the reaper out of this.
```
zramsay: i could have sworn a reviewed a logging PR very recently where `s/this/evts` ... any reason you're going back the other way?

logging/lifecycle/lifecycle.go
```diff
@@ -16,8 +15,11 @@ import (
 	"time"
```
zramsay: "time" should be up with "os", followed by \n, then any eris dependencies, then another \n then external dependencies

logging/loggers/eris_format_logger.go
```diff
@@ -0,0 +1,38 @@
+package loggers
+
+import (
+	"fmt"
+
+	"github.com/eris-ltd/eris-db/logging/structure"
+	kitlog "github.com/go-kit/kit/log"
```
zramsay: \n

event/event_cache_test.go
silasdavis: You did! The code usages are intact, but my refactor accidentally clobbered a bunch of usages of the English word 'this', which was unintentional, and which this corrects.

client/methods/call.go
silasdavis: Yeah I agree I was working off the precedent with the exit thing.

As an aside, I don't like `Warn`s; they seem like they;ll be useful but they end up being a pain in the arse. I'd rather force the code author to do something drastic (error or exit) or just provide an informational message and make sure continued execution is valid, because I think providing a warning option gives implied license to do unsafe things because 'they were warned', when a lot of the time warnings are ignored/unobserved. Also various packages will 'warn' over things that can never be resolved (because of your architecture for example) so I think the stock of warnings are is pretty low.

In this case, we can return errors from the `methods` package, and `Fatalf` them in the `cmd` package. Don't see any problem with this, but then again when I see `ClientDo` I tend to think we should be able to assume we are at interactive context (and so are allowed to exit). So that is perhaps a separate argument for evicting ClientDo from this level (I'll see how that looks).

client/cmd/transaction.go
silasdavis: Yeah I think this is legitimate within the cmd package

client/cmd/transaction.go
silasdavis: As above - I've remove these from methods - but as a helper in the cmd package I think exit here is apt

client/node_client.go
```diff
@@ -58,37 +58,31 @@ type NodeWebsocketClient interface {
...
-// NOTE [ben] Compiler check to ensure ErisNodeClient successfully implements
+// NOTE [ben] Compiler check to ensure erisNodeClient successfully implements
 // eris-db/client.NodeClient
-var _ NodeClient = (*ErisNodeClient)(nil)
+var _ NodeClient = (*erisNodeClient)(nil)
 
 // Eris-Client is a simple struct exposing the client rpc methods
-
-type ErisNodeClient struct {
+type erisNodeClient struct {
```
benjaminbollen: Should this not remain a public type?

client/websocket_client.go
```diff
@@ -41,29 +41,30 @@ type Confirmation struct {
...
 }
 
-// NOTE [ben] Compiler check to ensure ErisNodeClient successfully implements
+// NOTE [ben] Compiler check to ensure erisNodeClient successfully implements
 // eris-db/client.NodeClient
-var _ NodeWebsocketClient = (*ErisNodeWebsocketClient)(nil)
+var _ NodeWebsocketClient = (*erisNodeWebsocketClient)(nil)
 
-type ErisNodeWebsocketClient struct {
+type erisNodeWebsocketClient struct {
```
benjaminbollen: less important here, but equally I don't understand why this type is made private.

docs/generator/generator.go
```diff
@@ -0,0 +1,333 @@
+package generator
```
benjaminbollen: Good!  yes, happy to see the docs generator included here.

keys/key_client.go
```diff
@@ -29,31 +31,33 @@ type KeyClient interface {
...
 }
 
-// NOTE [ben] Compiler check to ensure ErisKeyClient successfully implements
+// NOTE [ben] Compiler check to ensure erisKeyClient successfully implements
 // eris-db/keys.KeyClient
-var _ KeyClient = (*ErisKeyClient)(nil)
+var _ KeyClient = (*erisKeyClient)(nil)
 
-type ErisKeyClient struct {
+type erisKeyClient struct {
```
benjaminbollen: Again, same question: not sure why these types are made private.

VoR0220: Yea this needs to be remade public for epm usage...

client/node_client.go
VoR0220: Yes. Needs to remain exported.

silasdavis: This was a hygiene thing. Note that the methods are on the interface `NodeClient`. Generally I avoid exporting the concrete struct type that implements and interfaces because I don't want people constructing it without using my constructor, `NewErisNodeClient` in this case. Bear in mind you don't need to export the struct to use the type via the exported constructor. I use this pattern a lot: https://github.com/eris-ltd/eris-db/blob/feature_logging/logging/loggers/info_trace_logger.go#L16.

In this case I particularly thing it should be unexported because it has private fields so an external package won't be able to initialise it properly for certain.

VoR0220: Ah...good point.

benjaminbollen: Good point, thanks.

PR #454, opened 2017-01-25
Glide dependency updating tool

silasdavis:
We have a situation with the way glide currently works that mean we really rely on hand-editing the glide.lock file to upgrade dependencies. However glide does provide most of what we need. This is an effort to solve a common use face while we wait for glide to integrate the more full-featured GPS library (which should allow our use cases).

Hell allows you to upgrade a package using glide and simply absorb that package's glide.lock file into your own (as well as upgrading glide.yaml for future use).

It does what we wish `glide get` would (for our case). That is, it leaves all other dependencies alone and just pulls in first-level transitive dependencies specified in a dependency's glide.lock file.

The usage is something like:

`go run ./hell/cmd/hell/main.go get --interactive github.com/tendermint/tendermint`                     

This will use glide as library to get a version of tendermint (you can choose the release) and then tendermint's glide.lock into our own. This way we obtain the transitive dependencies that any glide-using project has specified.

It's a bit of a blunt instrument in that it will just overwrite conflicting dependencies. But most of the time this is what we want. It should tide us over until principled full-featured package management/conflict resolution lands in glide.


VoR0220:
I love this name.

benjaminbollen:
Would it make sense to have this under `/util`, otherwise Im fine with it

PR #455, opened 2017-01-25
Feature logging

silasdavis:
Once we have https://github.com/eris-ltd/eris-db/pull/456

Suggest we merge the logging changes into develop. All of these changes have been reviewed.

The main thing missing at this stage is configuration which will be a subsequent pull request. Otherwise the logging works fine and people can base off this.

We _will_ have a slightly inconsistent logging output because we are not catching all of tendermint logs. Also all our logs will contain a rather long file-name trace, which will be configurable.


logging/adapters/adapters.go
```diff
@@ -0,0 +1,2 @@
+package adapters
+
```
zramsay: what's the file for?

logging/adapters/tendermint_log15/convert.go
```diff
@@ -0,0 +1,70 @@
+package adapters
+
+import (
+	"time"
+
+	"github.com/eris-ltd/eris-db/logging/loggers"
+	"github.com/eris-ltd/eris-db/logging/structure"
+	. "github.com/eris-ltd/eris-db/util/slice"
```
zramsay: we should avoid shorthanding unless there's a good reason

logging/loggers/channel_logger_test.go
```diff
@@ -0,0 +1,32 @@
+package loggers
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"fmt"
```
zramsay: should be on line 5

logging/adapters/adapters.go
silasdavis: Removed

logging/structure/structure.go
```diff
@@ -0,0 +1,146 @@
+package structure
+
+import (
+	"reflect"
+
+	. "github.com/eris-ltd/eris-db/util/slice"
```
zramsay: remove period?

logging/structure/structure_test.go
```diff
@@ -0,0 +1,36 @@
+package structure
+
+import (
+	"testing"
+
+	. "github.com/eris-ltd/eris-db/util/slice"
```
zramsay: ibid

logging/loggers/channel_logger_test.go
silasdavis: I'll make another PR to goimports everything

manager/manager.go
```diff
@@ -21,12 +21,13 @@ import (
 
 	events "github.com/tendermint/go-events"
 
-	log "github.com/eris-ltd/eris-logger"
-
 	config "github.com/eris-ltd/eris-db/config"
 	definitions "github.com/eris-ltd/eris-db/definitions"
```
zramsay: these don't need to be declared

rpc/tendermint/test/shared.go
```diff
@@ -21,6 +21,7 @@ import (
 
 	"path"
 
+	"github.com/eris-ltd/eris-db/logging/lifecycle"
 	state_types "github.com/eris-ltd/eris-db/manager/eris-mint/state/types"
 	"github.com/spf13/viper"
```
zramsay: external package should be on own line below

logging/adapters/tendermint_log15/convert.go
silasdavis: In general I would agree. In this case these functions are really intended for brevity and providing builtin-style functions for working with generic slices. I introduced them because there is a lot of that with the generic `Log(keyvals... interface{}) error` kitlog interface. I could be persuaded to step back, but I quite like it how it is.

logging/loggers/channel_logger_test.go
zramsay: i'll shutup then :+1:

manager/manager.go
silasdavis: Agreed. But not going to change that here.

rpc/tendermint/test/shared.go
silasdavis: Will goimports separately will touch may files so don't want to muddy this.

PR #461, opened 2017-01-27
Consolidation genesis make

benjaminbollen:


genesis/make_genesis_file.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/eris-ltd/eris-db/util"
+
+	"github.com/tendermint/go-crypto"
+	wire "github.com/tendermint/go-wire"
+)
+
+//------------------------------------------------------------------------------------
+// core functions
+
+func GenerateKnown(chainID, accountsPathCSV, validatorsPathCSV string) (string, error) {
```
silasdavis: Generate known as opposed to what? What is known?

genesis/make_genesis_file.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	pubkeys := make([]string, len(params))
+	amtS := make([]string, len(params))
+	names = make([]string, len(params))
+	permsS := make([]string, len(params))
+	setbitS := make([]string, len(params))
+	for i, each := range params {
+		pubkeys[i] = each[0]
+		amtS[i] = ifExistsElse(each, 1, "1000")
```
silasdavis: I could go either way, but I wonder if this would be nicer:

```go
ps := NewParamsWithDefaults(params, nil, "1000", "", fmt.Sprintf("%d", ptypes.DefaultPermFlags)), permsS[i])
...
ps.get(1)
```

Perhaps not worth it.

genesis/make_genesis_file.go
benjaminbollen: GenerateKnown is currently the signature of the function as it is used by cli.  So my plan here is to rewrite package genesis fully, but keep `GenerateKnown` so that we can work on the content of genesis within eris-db without blocking work on cli.  Then afterwards the name and signature of that function can be swapped out.  `GenerateKnown`, in my mind, refers to "known accounts written in a csv file"

genesis/make_genesis_file.go
benjaminbollen: While I was rewriting this, it will not come as a surprise that I simply got rid of "filling in randomly chosen parameters at randomly hardcoded places in the code".  The aim here is to get everything onto develop, so that cli can move forward.

PR #476, opened 2017-02-08
issue-476: return One256 for EXTCODESIZE on native contracts

VoR0220:
Fixes problem with not being able to get at the Snative contract. 

Signed-off-by: RJ Catalano <rj@monax.io>

### DIFF
--- a/manager/eris-mint/evm/vm.go
+++ b/manager/eris-mint/evm/vm.go
@@ -15,18 +15,19 @@ import (
 )
 
 var (
-	ErrUnknownAddress      = errors.New("Unknown address")
-	ErrInsufficientBalance = errors.New("Insufficient balance")
-	ErrInvalidJumpDest     = errors.New("Invalid jump dest")
-	ErrInsufficientGas     = errors.New("Insufficient gas")
-	ErrMemoryOutOfBounds   = errors.New("Memory out of bounds")
-	ErrCodeOutOfBounds     = errors.New("Code out of bounds")
-	ErrInputOutOfBounds    = errors.New("Input out of bounds")
-	ErrCallStackOverflow   = errors.New("Call stack overflow")
-	ErrCallStackUnderflow  = errors.New("Call stack underflow")
-	ErrDataStackOverflow   = errors.New("Data stack overflow")
-	ErrDataStackUnderflow  = errors.New("Data stack underflow")
-	ErrInvalidContract     = errors.New("Invalid contract")
+	ErrUnknownAddress         = errors.New("Unknown address")
+	ErrInsufficientBalance    = errors.New("Insufficient balance")
+	ErrInvalidJumpDest        = errors.New("Invalid jump dest")
+	ErrInsufficientGas        = errors.New("Insufficient gas")
+	ErrMemoryOutOfBounds      = errors.New("Memory out of bounds")
+	ErrCodeOutOfBounds        = errors.New("Code out of bounds")
+	ErrInputOutOfBounds       = errors.New("Input out of bounds")
+	ErrCallStackOverflow      = errors.New("Call stack overflow")
+	ErrCallStackUnderflow     = errors.New("Call stack underflow")
+	ErrDataStackOverflow      = errors.New("Data stack overflow")
+	ErrDataStackUnderflow     = errors.New("Data stack underflow")
+	ErrInvalidContract        = errors.New("Invalid contract")
+	ErrNativeContractCodeCopy = errors.New("Tried to copy native contract code, but this is not supported")
 )
 
 type ErrPermission struct {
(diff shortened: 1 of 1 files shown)


manager/eris-mint/evm/vm.go
```diff
@@ -565,13 +565,17 @@ func (vm *VM) call(caller, callee *Account, code, input []byte, value int64, gas
 			}
 			acc := vm.appState.GetAccount(addr)
 			if acc == nil {
-				return nil, firstErr(err, ErrUnknownAddress)
+				if _, ok := registeredNativeContracts[addr]; !ok {
```
benjaminbollen: Thus far we have not returned a code size for native contracts.  Is this motivated by a change in solidity, where it newly checks the code length before going to a contract.  We have not had a problem before calling our snative contracts.  Can you document and elaborate?

silasdavis:
Could you be a bit clearer with "problem with not being able to get at the Snative contract.", what problem prompted your investigation here?

It does seem like there can be a problem here and your solution will in some sense work. However this is a special case of a broader problem. That is making sure we have some special case handling for snatives. The other obvious place this will be an issue is EXTCODECOPY:

https://github.com/eris-ltd/eris-db/blob/master/manager/eris-mint/evm/vm.go#L583

in fact EXTCODESIZE seems like it probably is designed be used in tandem with that code for copy contract code around. We would need to handle some special case there too.

What is the context of use here? Pretending a native contract exists and has code length 1 may make certain code run, but it is ad hoc and it does nothing to allow you to treat native contracts as if they have code accessible to other contracts. It seems like the best thing to do here might be to just make in an error. We need more context.

Another place where we don't check whether we are dealing with an snative, and perhaps we should is in CREATE: https://github.com/eris-ltd/eris-db/blob/master/manager/eris-mint/evm/vm.go#L763

manager/eris-mint/evm/vm.go
```diff
@@ -565,13 +565,17 @@ func (vm *VM) call(caller, callee *Account, code, input []byte, value int64, gas
...
+					return nil, firstErr(err, ErrUnknownAddress)
+				}
+				stack.Push64(int64(1))
```
silasdavis: More idiomatic and a little more efficient to use `stack.Push(One256) here`

silasdavis:
Here's a fragment of solc 4.4.0 compiled bytecode that helps illuminate what the solidity compiler is up to:


```
      EXTCODESIZE 			perm.rm_role(addr, role)
      ISZERO 			perm.rm_role(addr, role)
      PUSH [ErrorTag] 			perm.rm_role(addr, role)
      JUMPI 			perm.rm_role(addr, role)
      PUSH 32			perm.rm_role(addr, role)
      GAS 			perm.rm_role(addr, role)
      SUB 			perm.rm_role(addr, role)
      CALL 			perm.rm_role(addr, role)
      ISZERO 			perm.rm_role(addr, role)
      PUSH [ErrorTag] 			perm.rm_role(addr, role)
      JUMPI 			perm.rm_role(addr, role)
      POP 			perm.rm_role(addr, role)
      POP 			perm.rm_role(addr, role)
      POP 			perm.rm_role(addr, role)
      PUSH 40			perm.rm_role(addr, role)
```

Here's the relevant line in the solidity compiler that generates this type of code: https://github.com/ethereum/solidity/blob/develop/libsolidity/codegen/ExpressionCompiler.cpp#L1662

The check amounts to an existence check for the contract/contract code before running it. I am not certain why this is, but it seems like in the current pricing model it might be better to check the code exists before calling empty code. Worth checking.

In this compiler case returning any truthy value seems to be the right thing to do, and 1 is probably the most elegant.

There is a worry with how this interacts with `EXTCODECOPY`. It could easily try and copy non-existing bytes into allocated space. However as a point of reassurance, a quick grep of the solidity compiler code does not seem to show any use of `EXTCODECOPY` by the the solidity code generator.

If someone is deliberately trying to access snative code from an `EXTCODECOPY` then we have a problem, but an error is probably the right thing to do there from the time being. We _could_ return stub code, but that seems like it would cause more pain than it might prevent. `EXTCODECOPY` is a low-level thing, and I think solidity is our main concern...

manager/eris-mint/evm/vm.go
```diff
@@ -565,21 +566,28 @@ func (vm *VM) call(caller, callee *Account, code, input []byte, value int64, gas
...
-				return nil, firstErr(err, ErrUnknownAddress)
+				if _, ok := registeredNativeContracts[addr]; !ok {
+					return nil, firstErr(err, ErrUnknownAddress)
+				}
+				stack.Push64(int64(One256))
+				dbg.Printf(" => Hit native contract\n\n")
+			} else {
+				code := acc.Code
+				l := int64(len(code))
+				stack.Push64(l)
```
silasdavis: again, I think this should be `stack.Push(One256)` it saves doing the conversion to use the constant, and it is used elsewhere

VoR0220: This is the actual code length for normal contracts that are not natives.

VoR0220: It should stay as it is.

benjaminbollen:
LGTM

Issue #417, opened 2016-12-07
Update keys in sNatives PermsMap to permit/improve sNatives access from Solidity


jasonklein:
The keys in `PermsMap` used to identify the sNatives functions at https://github.com/eris-ltd/eris-db/blob/master/manager/eris-mint/evm/snative.go#L55 are the signatures of functions identical to those in the `Permissions` contract at https://github.com/eris-ltd/eris-pm/blob/master/tests/fixtures/app33-evm_connection_edb_permissions_and_roles_layer/permissions.sol.

Please update the keys to represent (with respect to Solidity):
- a `set_global` function that takes 2 arguments, not three: `function set_global(uint64 permFlag, bool value) constant returns (uint64 pf)`
  - 🚨  this is the most immediately requested change because the inconsistency appears to render `set_global` unreachable from a Solidity contract
- functions that set the type for the `permFlag` argument to `uint64` rather than `int` (to be consistent with `PermFlag` in https://github.com/eris-ltd/eris-db/blob/master/permission/types/permissions.go#L17)
- functions with names that are camelCase rather than snake_case
  - `has_base` -> `hasBase`
  - `set_base` -> `setBase`
  - `unset_base` -> `unsetBase`
  - `set_global` -> `setGlobal`
  - `has_role` -> `hasRole`
  - `add_role` -> `addRole`
  - `rm_role` -> `rmRole`

VoR0220:
fairly sure I've completed this but it is a breaking change...cool if I just put the PR up in a "hold onto this until 0.17.0" branch?

silasdavis:
We need to to introduce the permissions stub contract as a first-class asset in eris-db, and automate it so we can be more certain the snatives permissions map matches the contract. It's not good that the permissions map is just some alien reference.

The primary issue here is the inaccessibility of `set_global`. I am reluctant to make cosmetic changes to the casing of the function names that will break backward compatibility, is there a strong reason to do this?

You are correct on the `uint64` point, but again I am reluctant to change the signatures of the working methods for little benefit. Bear in mind that since a PermFlag bitfield is really a way of encoding a more complex type into an integer being slightly tighter on the outer type seems to buy us rather less than it would if we were representing a strictly non-negative number.

[...]

PR #480, opened 2017-02-08
Function signatures automated and fix set global

VoR0220:
Solves set_global portion and uint consistency on #417 . Also automates function identifier generation for a signature.

### DIFF
--- a/manager/eris-mint/evm/snative.go
+++ b/manager/eris-mint/evm/snative.go
@@ -4,7 +4,9 @@ import (
 	"encoding/hex"
 	"fmt"
 
+	"github.com/eris-ltd/eris-db/manager/eris-mint/evm/sha3"
 	ptypes "github.com/eris-ltd/eris-db/permission/types"
+
 	. "github.com/tendermint/go-common"
 )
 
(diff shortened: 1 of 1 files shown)


silasdavis:
This looks good, but I'm thinking it would be nice to have a test closer to this file that would reassure as about changes. Can you link to an appropriate epm test?

VoR0220:
@silasdavis certainly I can do that....once I make it. :P I'll make an update to the permissions test.

benjaminbollen:
@silasdavis writing the unit tests for this was also still on my todo; happy for you to take this if you want.

silasdavis:
Let's merge this for now. I've made an epic for further improvements: https://github.com/eris-ltd/eris-db/issues/496.

Unit tests are nice, but I think really its the cotnract-level/EPM tests that are going to give us most certainty.

@VoR0220 do I understand correctly that you plan to make a test for these in eris?

VoR0220:
@silasdavis you are correct, I am working on one. But before that, I would like to know WHAT it is I am testing...because currently I have a vague idea of what setGlobal does.

silasdavis:
Ah, so we do have some kind of tests for this here: https://github.com/eris-ltd/eris-db/blob/master/manager/eris-mint/state/permissions_test.go

I wonder why they didn't detect this change... will have to look into.

VoR0220:
Because there wasnt a massive change...i just automated the func signatures. And we also dont test setGlobal currently.

silasdavis:
No, actually we do test `set_global` here (in permission_test.go), that was the point of my link.

VoR0220:
Ah crap missed that link...thought you were pointing at PM test. My bad.

benjaminbollen:
To be clear: I asked for a unit test that had the hard-coded signatures in place; as a breaker for accidentally changing the signature, and as such the function identifier.

PR #484, opened 2017-02-09
Logging config and outputs

silasdavis:
Some refinement still on-going and writing some additional tests, but putting out for review now.

I need to tweak default configs for client and server and then verify a few things around capture.

Possibly we should provide some default verbosity filters for finer grain control than info and trace, but the LoggingConfig mechanism is pretty general.


glide.lock
```diff
@@ -42,37 +52,45 @@ imports:
   version: d9eb7a3d35ec988b8585d4a0068e462c27d28380
 - name: github.com/gorilla/websocket
   version: a68708917c6a4f06314ab4e52493cc61359c9d42
+- name: github.com/Graylog2/go-gelf
+  version: 5bfd5bbbfe86489017c268bf86539fcec7e28d8e
 - name: github.com/hashicorp/hcl
```
benjaminbollen: informational question: do we know where hashicorp/hcl sits in the dependency tree?

logging/config/config.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (lc *LoggingConfig) TOMLString() string {
+	return tomlString(lc)
+}
+
+func LoggingConfigFromMap(loggingRootMap map[string]interface{}) (*LoggingConfig, error) {
+	lc := new(LoggingConfig)
+	buf := new(bytes.Buffer)
+	enc := toml.NewEncoder(buf)
+	// TODO: [Silas] consider using strongly typed config/struct mapping everywhere
```
benjaminbollen: Noted.

logging/config/config.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	return lc, nil
+}
+
+func tomlString(v interface{}) string {
+	buf := new(bytes.Buffer)
+	encoder := toml.NewEncoder(buf)
+	err := encoder.Encode(v)
+	if err != nil {
+		// Seems like a reasonable compromise to make the string function clean
```
benjaminbollen: **shivers** because it is in logging/config, but I'm not fully on-board here. Ok to pass.

logging/config/sinks_test.go
```diff
@@ -0,0 +1,112 @@
...
+			Sink().
+				AddSinks(
+					Sink().
+						AddSinks(
+							Sink().
+								SetTransform(CaptureTransform("cap", 100, true)).
+								SetOutput(StderrOutput()).
+								AddSinks(
+									Sink().
+										SetTransform(LabelTransform(true, "Label", "A Label!")).
```
benjaminbollen: Label: A Label!

glide.lock
silasdavis: Pulled in by Viper (util.go)

logging/config/config.go
silasdavis: The reasoning is, this will generate pretty obviously failing config if it does slip through, shouldn't cause a runtime issue, and primarily because failures are highly likely to be caught by tests. It doesn't seem reasonable to error or panic on encoding a well-known type like this.

That said, I also feel borderline about this.

logging/config/sinks_test.go
silasdavis: It's a label, if that was unclear

benjaminbollen:
So if you can remove the eris-keys from glide; then Im happy to merge this

PR #499, opened 2017-02-15
Cleanup: introduce word256 and remove dependency on tendermint/go-common

benjaminbollen:
While working on release-0.16, to have the eris-db definitions for Permissions to be a clean import into eris tooling, I removed the lingering dependency on `tendermint/go-common`; this saw me forced to pull this new type through the EVM and state as well; and finally there was then no need to preserve `tendermint/go-common` for sanity checks and random numbers for tests.

- introduces `eris-db/word256` for bytes32 and padding and conversion functions
- introduces `eris-db/common/sanity` for wrapping `PanicSanity` and `PanicCrisis`
- introduces `eris-db/common/random` for wrapping around `rand` and `crypto/rand`

- changes imports across `eris-db/manager` and `eris-db/permissions`

### DIFF
--- a/account/account.go
+++ b/account/account.go
@@ -25,8 +25,9 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/eris-ltd/eris-db/common/sanity"
 	ptypes "github.com/eris-ltd/eris-db/permission/types"
-	. "github.com/tendermint/go-common"
+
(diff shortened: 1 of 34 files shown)


benjaminbollen:
Yes, I agree, particularly by separating it into a `word256` instead of a bigger `common` bag; it makes it very clear what the unit testing for it should be

PR #500, opened 2017-02-17
Introduce more strongly typed SNatives and contract templates

### LINKED ISSUE #417: Update keys in sNatives PermsMap to permit/improve sNatives access from Solidity

### LINKED ISSUE #418: [SNative] Don't return nil from SNative Dispatch
`permissionsContract` in https://github.com/eris-ltd/eris-db/blob/master/manager/eris-mint/evm/snative.go#L65 returns `nil` (among other return values) if the `caller` does not have the invoked permission.

The Solidity functions that are used to access the sNatives functions (see `Permissions`, https://github.com/eris-ltd/eris-pm/blob/master/tests/fixtures/app33-evm_connection_edb_permissions_and_roles_layer/permissions.sol) expect to receive `int`s (ideally changed to `uint64`s, see https://github.com/eris-ltd/eris-db/issues/417) or `bool`, depending on the function. Receiving `nil` gets `Invalid jump dest`.

Can `permissionsContract` be changed to return the default value of the permission intended to be invoked (so, either `0` or `false`, I think)?

### LINKED ISSUE #498: [SNatives] Document permissions SNatives
Add documentation in snative.go.

This will inform the integration-level tests that @VoR0220 will write for: https://github.com/eris-ltd/eris/issues/1227. Once this is done, give that ticket a bump.

silasdavis:
This PR aims to address the issue of having many versions of the permissions contract in different places. Since eris-db is responsible for interpreting SNative calls it seems sensible that eris-db holds the defintion for SNatives. This PR:

- Introduces a more strongly typed (the functions themselves are still not strongly typed) SNative and contract definition
- Allows you to generate a permissions contract with comments that compiles and can be exported to other projects like smart contract bundles
- Adds a test for the permissions contract

It doesn't change the fundamentals of how SNatives work.

The permissions contract structure is built from the libraries' team's version: https://github.com/eris-ltd/eris-contract-bundles/blob/master/src/commons-auth/contracts/SecureNativeAuthorizations.sol which Jason linked me too which is nicely commented.

Outstanding for this is:
- [x] Finalise what types we are going to use for the permissions function
- [x] Possibly take opportunity to change the naming conventions as in https://github.com/eris-ltd/eris-db/issues/417
- [x] Add a couple more unit tests

Once that is done then we can mechanical generate the solidity interface that we are advertising to clients and we can do this for any future SNatives (which is why I thought it was worth doing). By running:

```go
vm.SNativeContracts()["permissions_contract"].Solidity()
```

Also note that the current version of snative.go on release-0.16 is missing a bracket in the signature.

fixes #498 
fixes #418 
fixes #417


VoR0220:
Am in favor of changing the names. So yea...lets make this happen.

silasdavis:
@VoR0220  Can we get some consensus on what the type signatures should actually be also?

VoR0220:
@silasdavis i think i know what you mean but youre going to need to clarify. I think you mean solidity types. In which case there should be changes. For one, all ints should be uint64s to better align with golang. We should also contemplate switching from bytes32 to string as it could enable longer strings and dynamic ones at that.

silasdavis:
What I need is a decision on:

- int or uint64 (sounds like uint64)
- bytes32 or string (sounds like bytes32)
- function name casing (sounds like camel case)
- The name of contract (and therefore its address)

These will be breaking changes, but seeing as it was already not working properly I think we are at liberty to make them if they improve the situation.

Here is the contract as currently defined in snatives.go: 

```javascript
contract permissions_contract {

/**
* @notice Sets global (default) value for a base authorization
* @param _account account
* @param _authorization base authorization
* @param _value value of base authorization
* @return authorization base authorization passed in
*/
function set_global(address _account, int _authorization, int _value) constant returns (int authorization);

/**
* @notice Adds a role to an account
* @param _account account
* @param _role role
* @return result whether role was added
*/
function add_role(address _account, bytes32 _role) constant returns (bool result);

/**
* @notice Indicates whether an account has a role
* @param _account account
* @param _role role
* @return result whether account has role
*/
function has_role(address _account, bytes32 _role) constant returns (bool result);

/**
* @notice Removes a role from an account
* @param _account account
* @param _role role
* @return result whether role was removed
*/
function rm_role(address _account, bytes32 _role) constant returns (bool result);

/**
* @notice Sets a base authorization for an account
* @param _account account
* @param _authorization base authorization
* @param _value value of base authorization
* @return result value passed in
*/
function set_base(address _account, int _authorization, int _value) constant returns (bool result);

/**
* @notice Indicates whether an account has a base authorization
* @param _account account
* @param _authorization base authorization
* @return result whether account has base authorization set
*/
function has_base(address _account, int _authorization) constant returns (bool result);

/**
* @notice Sets a base authorization for an account to the global (default) value of the base authorization
* @param _account account
* @param _authorization base authorization
* @return authorization base authorization passed in
*/
function unset_base(address _account, int _authorization) constant returns (int authorization);

}
```

VoR0220:
let me play with turning it to string and get back to you later today. It may very well be worth it to send in a string instead of a bytes32 simply because a string can be longer than 32 bytes (why you would ever make a string like that, IDK, but the liberty is nice).

manager/eris-mint/evm/snative_test.go
```diff
@@ -0,0 +1,143 @@
...
+3f0ebb30 set_base(address,uint64,uint64)
+67dc6f70 set_global(address,uint64,uint64)
+73448c99 unset_base(address,uint64)
+*/
+
+func TestPermissionsContractSignatures(t *testing.T) {
+	contract := SNativeContracts()["permissions_contract"]
+
+	assertContractFunction(t, contract, "3fbf7da5",
+		"add_role(address,bytes32)")
```
VoR0220: Thought we said camelCase?

silasdavis: Ah right yeah, I can do that, but it will involve a few more changes. Obviously just a search and replace job in Eris db so I can do it, but it got me to question the wisdom, will do it if you and Jason and Ben think it's good idea

manager/eris-mint/evm/snative.go
```diff
@@ -1,104 +1,351 @@
...
+//-----------------------------------------------------------------------------
+// snative are native contracts that can access and modify an account's permissions
+
+func NewSNativeContract(comment, name string, functions ...SNativeFuncDescription) SNativeContractDescription {
+	fs := make(map[FuncID]SNativeFuncDescription, len(functions))
+	for _, f := range functions {
+		fid := f.ID()
+		otherF, ok := fs[fid]
+		if ok {
+			panic(fmt.Errorf("Function with ID %x already defined: %s", fid,
```
benjaminbollen: This should not panic, even if it would only be called on initialization;

manager/eris-mint/evm/snative.go
```diff
@@ -1,104 +1,351 @@
...
+		return Zero256.Bytes(), err
+	}
+
+	remainingArgs := args[FuncIDLength:]
 
 	// check if we have permission to call this function
-	if !HasPermission(appState, caller, d.PermFlag) {
-		return nil, ErrInvalidPermission{caller.Address, d.Name}
+	if !HasPermission(appState, caller, function.PermFlag) {
+		return Zero256.Bytes(), ErrInvalidPermission{caller.Address, function.Name}
```
benjaminbollen: Good on returning `Zero256`

manager/eris-mint/evm/snative.go
```diff
@@ -1,104 +1,351 @@
...
+	}
 }
 
-func permissionsContract(appState AppState, caller *Account, args []byte, gas *int64) (output []byte, err error) {
-	if len(args) < 4 {
-		return nil, fmt.Errorf("permissionsContract expects at least a 4-byte function identifier")
+// Returns a map of all SNative contracts defined indexed by name
+func SNativeContracts() map[string]SNativeContractDescription {
+	permFlagType := SolidityUint64
+	roleType := SolidityBytes32
```
benjaminbollen: Internally we treat roles as strings, but I agree that it is better and sufficient to type it strictly as `bytes32`; alternatively we should change our impl to treat it as `Word256`

manager/eris-mint/evm/snative.go
```diff
@@ -1,104 +1,351 @@
...
+type SNativeContractDescription struct {
+	Comment   string
+	Name      string
+	functions map[FuncID]SNativeFuncDescription
 }
 
-//-----------------------------------------------------------------------------
-// snative are native contracts that can access and modify an account's permissions
-
 type SNativeFuncDescription struct {
```
benjaminbollen: Second thought / correction: `SNativeFuncDescription` should not become an interface as suggested above.  Rather the SNative definition should be defined with `LanguageTypes` and the templating should be a package on itself, that produces the solidity code based on the mapping of `LanguageType` to `SolidityType`.  We can raise this as an issue and proceed as is for now; the net gain in readability is sufficient for me.

benjaminbollen: Make that `ProtoType` rather than `LanguageType`

benjaminbollen:
LGTM.  I have raised comments on the unpleasantness of infection of solidity into the evm/snatives; Talked with Silas and he will examine further; although he does not like to call it `ProtoTypes` :)

silasdavis:
The latest commit:

- Refactors Solidity-specific code into a util package and introduces a templating wrapper, and some formatting improvements (like proper indenting)
- Introduces an `abi` package under `evm` to decouple Solidity types and calling conventions from those defined by Ethereum by way of the ABI (this can be one path to future expansion to support an ABI-aware chain more generally)
- Improves documentation and naming of variables on contract
- Changes the 'value' argument on `set_*` functions to be a bool as it should be
- Changes the return value from `*set_*` functions to be the resultant permission after the permission operation has applied (rather than echoing back the argument the caller passed in). This retains the primary purpose of the return value as a truthy call success flag.

Makefile
```diff
@@ -73,6 +73,11 @@ hell:
 	go build -o ${REPO}/target/hell ./util/hell/cmd/hell/main.go
 	./target/hell $(filter-out $@,$(MAKECMDGOALS))
 
+# Dumps Solidity interface contracts for SNatives
+.PHONY: snatives
+snatives:
+	@go run ./util/snatives/cmd/main.go
```
VoR0220: Tried this out, works like a charm. Good work.

event/event_cache_test.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-var mockInterval = 10 * time.Millisecond
+var mockInterval = 40 * time.Millisecond
```
VoR0220: Reason for this?

manager/eris-mint/evm/abi/types.go
```diff
@@ -0,0 +1,27 @@
+package abi
```
VoR0220: Should clarify in the notes below that this is an ABI for our SNative contracts.

manager/eris-mint/evm/fake_app_state.go
```diff
@@ -3,7 +3,6 @@ package vm
 import (
 	"fmt"
 
-	. "github.com/eris-ltd/eris-db/manager/eris-mint/evm"
```
VoR0220: ++

manager/eris-mint/evm/log_event_test.go
```diff
@@ -5,7 +5,6 @@ import (
 	"reflect"
 	"testing"
 
-	. "github.com/eris-ltd/eris-db/manager/eris-mint/evm"
```
VoR0220: ++

manager/eris-mint/evm/snative.go
```diff
@@ -1,104 +1,286 @@
...
-		// so calling them would be unsafe ...
-		registeredNativeContracts[LeftPadWord256([]byte("has_base"))] = has_base
-		registeredNativeContracts[LeftPadWord256([]byte("set_base"))] = set_base
-		registeredNativeContracts[LeftPadWord256([]byte("unset_base"))] = unset_base
-		registeredNativeContracts[LeftPadWord256([]byte("set_global"))] = set_global
-		registeredNativeContracts[LeftPadWord256([]byte("has_role"))] = has_role
-		registeredNativeContracts[LeftPadWord256([]byte("add_role"))] = add_role
-		registeredNativeContracts[LeftPadWord256([]byte("rm_role"))] = rm_role
-	*/
+type SNativeContractDescription struct {
```
VoR0220: Needs Godoc documentation of what each of the parameters that are exported do.

manager/eris-mint/evm/snative.go
```diff
@@ -1,104 +1,286 @@
...
+	Comment   string
+	Name      string
+	functions map[FuncID]*SNativeFunctionDescription
 }
 
-//-----------------------------------------------------------------------------
-// snative are native contracts that can access and modify an account's permissions
-
-type SNativeFuncDescription struct {
+type SNativeFunctionDescription struct {
```
VoR0220: Same here.

manager/eris-mint/evm/snative.go
```diff
@@ -115,7 +297,7 @@ func has_base(appState AppState, caller *Account, args []byte, gas *int64) (outp
 }
 
 func set_base(appState AppState, caller *Account, args []byte, gas *int64) (output []byte, err error) {
-	addr, permNum, perm := returnThreeArgs(args)
+	addr, permNum, permVal := returnThreeArgs(args)
```
VoR0220: much better name ++

manager/eris-mint/evm/snative.go
```diff
@@ -228,6 +410,10 @@ func ValidPermN(n ptypes.PermFlag) bool {
 	return true
 }
 
+func resultantPermBytes(basePerms ptypes.BasePermissions) []byte {
```
VoR0220: very nice.

manager/eris-mint/evm/snative_test.go
```diff
@@ -0,0 +1,124 @@
...
+	. "github.com/eris-ltd/eris-db/manager/eris-mint/evm/opcodes"
+	ptypes "github.com/eris-ltd/eris-db/permission/types"
+)
+
+/* Compiling the Permissions solidity contract at
+(generated by Solidity() function)
+https://ethereum.github.io/browser-solidity yields:
+
+Functions
+3fbf7da5 add_role(address,bytes32)
```
VoR0220: Need to decide if we're going CamelCase Here...or mixedCase (as is Solidity convention).

manager/eris-mint/evm/snative.go
```diff
@@ -1,104 +1,286 @@
...
+// Returns a map of all SNative contracts defined indexed by name
+func SNativeContracts() map[string]*SNativeContractDescription {
+	permFlagType := abi.Uint64
+	roleType := abi.Bytes32
+	contracts := []*SNativeContractDescription{
+		NewSNativeContract(`
+		* Interface for managing Secure Native authorizations.
+		* @dev This interface describes the functions exposed by the SNative permissions layer in the Monax blockchain (ErisDB).
+		`,
+			"permissions_contract",
```
VoR0220: permissions_contract or are we doing PermissionsContract? Either way I'm fine. Just clarifying.

event/event_cache_test.go
silasdavis: I had a CI failure, timing here was a little tight anyway, also ugly, but pre-existing and we'll get around to that.

VoR0220:
Very approved. Merge away.

silasdavis:
In for a penny, in for pound. Have changed case convention of permissions functions to camelCase. Also renamed permission contract to `Permissions` so the canonical address for it will have changed. Also automated testing a bit more.

The generated solidity for the permissions contract now looks like this:

```javascript
/**
* Interface for managing Secure Native authorizations.
* @dev This interface describes the functions exposed by the SNative permissions layer in the Monax blockchain (ErisDB).
* @dev These functions can be accessed as if this contract were deployed at the address 0x0000000000000000005065726d697373696f6e73
*/
contract Permissions {
    /**
    * @notice Adds a role to an account
    * @param _account account address
    * @param _role role name
    * @return result whether role was added
    */
    function addRole(address _account, bytes32 _role) constant returns (bool result);

    /**
    * @notice Removes a role from an account
    * @param _account account address
    * @param _role role name
    * @return result whether role was removed
    */
    function removeRole(address _account, bytes32 _role) constant returns (bool result);

    /**
    * @notice Indicates whether an account has a role
    * @param _account account address
    * @param _role role name
    * @return result whether account has role
    */
    function hasRole(address _account, bytes32 _role) constant returns (bool result);

    /**
    * @notice Sets the permission flags for an account. Makes them explicitly set (on or off).
    * @param _account account address
    * @param _permission the base permissions flags to set for the account
    * @param _set whether to set or unset the permissions flags at the account level
    * @return result the resultant permissions flags on the account after the call
    */
    function setBase(address _account, uint64 _permission, bool _set) constant returns (uint64 result);

    /**
    * @notice Unsets the permissions flags for an account. Causes permissions being unset to fall through to global permissions.
    * @param _account account address
    * @param _permission the permissions flags to unset for the account
    * @return result the resultant permissions flags on the account after the call
    */
    function unsetBase(address _account, uint64 _permission) constant returns (uint64 result);

    /**
    * @notice Indicates whether an account has a subset of permissions set
    * @param _account account address
    * @param _permission the permissions flags (mask) to check whether enabled against base permissions for the account
    * @return result whether account has the passed permissions flags set
    */
    function hasBase(address _account, uint64 _permission) constant returns (uint64 result);

    /**
    * @notice Sets the global (default) permissions flags for the entire chain
    * @param _permission the permissions flags to set
    * @param _set whether to set (or unset) the permissions flags
    * @return result the resultant permissions flags on the account after the call
    */
    function setGlobal(uint64 _permission, bool _set) constant returns (uint64 result);
}
```

manager/eris-mint/state/tx_cache.go
```diff
@@ -19,6 +19,8 @@ type TxCache struct {
 	storages map[Tuple256]Word256
 }
 
+var _ vm.AppState = &TxCache{}
```
benjaminbollen: A comment is not unwelcome; but great! thx for adding this line

permission/types/permissions.go
```diff
@@ -112,6 +112,12 @@ func (p *BasePermissions) IsSet(ty PermFlag) bool {
 	return p.SetBit&ty > 0
 }
 
+// Returns the Perms PermFlag masked with SetBit bit field to give the resultant
+// permissions enabled by this BasePermissions
+func (p *BasePermissions) ResultantPerms() PermFlag {
```
benjaminbollen: this is not ok as a public function;  It can encourage usage of it and it should be clear that effective permissions are those of the global permissions + delta of base permissions for that account. Why would we have this function?

util/snatives/templates/solidity_templates.go
```diff
@@ -0,0 +1,133 @@
+package templates
```
benjaminbollen: Nice ! I like that you moved them to util; even better than having them in evm

client/rpc/client_test.go
```diff
@@ -147,7 +147,7 @@ func testPermissions(t *testing.T,
 	nonceString := ""
 
 	_, err := Permissions(nodeClient, keyClient, publicKeyString, addressString,
-		nonceString, "set_base", []string{permAddressString, "root", "true"})
+		nonceString, "setBase", []string{permAddressString, "root", "true"})
```
VoR0220: If we're going to harden these, now would be the time to add a test for every permission. This way we can effectively gauge what parts are on the jobs failing and what parts are on e-db failing, atleast more easily...currently I'm having trouble deciphering this.

benjaminbollen: There are extensive unit tests in `/manager/eris-mint/state/permissions_test.go` that have been passing without failure

VoR0220: That's fine, but working directly on the EVM/SNative is not the same as going through the RPC (as we've seen with https://github.com/eris-ltd/eris-db/pull/477). I would recommend that there is atleast one of every RPC permission call type included in here and to ensure that they are passing.

benjaminbollen:
LGTM

manager/eris-mint/state/tx_cache.go
silasdavis: "TxCache implements vm.AppState", what sort of comment were you after?

permission/types/permissions.go
silasdavis: If you want to see what permissions are set at the account level. Knowing effective permissions is also useful I agree.

manager/eris-mint/state/tx_cache.go
benjaminbollen: optionally yes, it s not entirely obvious otherwise

permission/types/permissions.go
silasdavis: Given that `BasePermissions.Perms` and `BasePermissions.SetBit` are both public fields I'd say this was arguably less obscure in meaning than those. Callers are allowed to interrogate account level permissions and global permissions separately.

PR #501, opened 2017-02-19
Release 0.16 - alignment with consolidated tooling wrt chain making

benjaminbollen:
After consolidation of eris-cm and eris-pm into eris tooling; some technical debt had to be resolved.  This PR is necessary for this re-alignment in eris https://github.com/eris-ltd/eris/pull/1235 to complete.

checklist:
- [x] clean up eris-db dependencies of `permissions`; merged in #499 
- [x] permissions utility to convert map[string]int into permissions type
- [x] genesis: define constructors for Genesis types to contain explicit dependency on tendermint/go-crypto into tooling; enabling us to inject our own crypto package without coordination with tooling.
- [x] genesis: introduce GenesisPrivateValidtor as an intermediary step to removing priv_validator logic from the stack
- [x] genesis: provide `genesis.GenerateGenesisFileBytes(chainName, []*GenesisAccount, []*GenesisValidator)` for tooling
- [x] final review


genesis/make_genesis_file.go
```diff
@@ -28,6 +28,25 @@ func GenerateKnown(chainID, accountsPathCSV, validatorsPathCSV string) (string,
 }
 
 //------------------------------------------------------------------------------------
+// interface functions that are consumed by monax tooling
+
+func GenerateGenesisFileBytes(chainName string, genesisAccounts []*GenesisAccount,
```
silasdavis: Similar to other comment, if it's convenient to pass `genesisAccounts []GenesisAccount` then we could probably get away without a clone... (though we do have pointers from `GenesisAccount` to `AccountPermissions`, but I don't think we accidentally mutate here)

genesis/maker.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func NewGenesisValidator(amount int64, name string, unbondToAddress []byte,
+	unbondAmount int64, keyType string, publicKeyBytes []byte) (*GenesisValidator, error) {
+	// convert the key bytes into a typed fixed size byte array
+	var typedPublicKeyBytes []byte
+	switch keyType {
+	case "ed25519":
+		// TODO: [ben] functionality and checks need to be inherit in the type
+		if len(publicKeyBytes) != 32 {
```
silasdavis: I would rather replace all magic number references with named constants

genesis/maker.go
```diff
@@ -0,0 +1,80 @@
...
+			return nil, fmt.Errorf("Invalid length provided for ed25519 public key (len %v)",
+				len(publicKeyBytes))
+		}
+		// ed25519 has type byte 0x01
+		typedPublicKeyBytes = make([]byte, 33)
```
silasdavis: likewise would prefer `publicKeyLength + 1`

genesis/maker.go
```diff
@@ -0,0 +1,80 @@
...
+		// prepend type byte to public key
+		typedPublicKeyBytes = append([]byte{crypto.PubKeyTypeEd25519}, publicKeyBytes...)
+	case "secp256k1":
+		if len(publicKeyBytes) != 64 {
```
silasdavis: again

genesis/maker.go
```diff
@@ -0,0 +1,80 @@
...
+			return nil, fmt.Errorf("Invalid length provided for secp256k1 public key (len %v)",
+				len(publicKeyBytes))
+		}
+		// secp256k1 has type byte 0x02
+		typedPublicKeyBytes = make([]byte, 65)
```
silasdavis: same

genesis/types.go
```diff
@@ -65,3 +94,71 @@ func GenesisDocFromJSON(jsonBlob []byte) (genState *GenesisDoc) {
...
+	if genesisValidator == nil {
+		return GenesisValidator{}, fmt.Errorf("Cannot clone nil GenesisValidator.")
+	}
+	if genesisValidator.PubKey == nil {
+		return GenesisValidator{}, fmt.Errorf("Invalid GenesisValidator %s with nil public key.",
+			genesisValidator.Name)
+	}
+	// clone the public key by writing and reading over go-wire serialisation
+	// TODO! write unit test to see whether this is correct
+	publicKeyClone, err := crypto.PubKeyFromBytes(genesisValidator.PubKey.Bytes())
```
silasdavis: I can see how this is sort of defensive. But I wonder if it is a step too far. Ultimately we have to trust that  we are not being handed internal bytes slice that we might mutate. We can see for sure that we are not being handed a reference to internal bytes as it is.

genesis/genesis.go
```diff
@@ -0,0 +1,85 @@
...
+// MakeGenesisDocFromAccounts takes a chainName and a slice of pointers to GenesisAccount,
+// and a slice of pointers to GenesisValidator to construct a GenesisDoc, or returns an error on
+// failure.  In particular MakeGenesisDocFromAccount uses the local time as a
+// timestamp for the GenesisDoc.
+func MakeGenesisDocFromAccounts(chainName string, accounts []*GenesisAccount,
+	validators []*GenesisValidator) (GenesisDoc, error) {
+
+	// TODO: assert valid accounts and validators
+	// TODO: [ben] expose setting global permissions
+	globalPermissions := ptypes.DefaultAccountPermissions.Clone()
```
benjaminbollen: Yes, but is a struct `AccountPermissions` of a struct `BasePermissions` and a slice of `strings`; I believe - but not sure - that the struct `BasePermissions` is treated as a pointer essentially; but definitely the `Roles []string` is an effective pointer;

secondly, genesis is currently constructed with `Params` as a pointer to a `GenesisParams` which is a pointer to an `AccountPermissions`;  I dont particularly like this construction; but I didnt want to end up with a `GenesisDoc` that didnt own its own data;

Arguably this does read very defensive

genesis/make_genesis_file.go
benjaminbollen: in tooling there is a usage of passing slices of pointers; which is justified there because they are passed around a lot during construction;

this is why i deliberately gave tooling an interface that respects the `[ ]*T`; but `GenesisDoc` is constructed with `[ ]T` and once `GenesisDoc` is constructed changing the original `T` should not affect the `GenesisDoc`

genesis/maker.go
benjaminbollen: yes, but `tendermint/go-crypto/pub_key.go` does not have this definition; and while I agree with you.  I prefer to hold off making that definition until (very soon) we make `eris-db/crypto`

genesis/maker.go
benjaminbollen: yes, on the same page, but without a `const`, `32+1` is a bit weird

genesis/types.go
benjaminbollen: Having a `Clone` in general is very good practice, because it allows you to make it explicit when you're copying data; and as a result avoid copying it when you don't have to.

Here again Clone is introduced, because on construction of the genesisDoc I want to have unique ownership of GenesisDoc

genesis/genesis.go
silasdavis: The inner struct is just another value, the memory is copied in a single pass. See: https://play.golang.org/p/Ut_7gZ3TyJ

Yeah the slice does give you mutable access. But this is go, I think we need to choose our battles. I'd rather we were passing immutable sequences of values but we're not. And in the event we are not writing back to the roles, and that would be a weird things to do.

genesis/make_genesis_file.go
silasdavis: thought that might be the case. Makes sense.

genesis/maker.go
silasdavis: yep, which is why you should introduce the constant. In the same file if necessary.

genesis/maker.go
silasdavis: We should just name it in the same file, let's not make the perfect an enemy of the good.

genesis/types.go
silasdavis: I do get it, it's just using non-pointer values is idiomatic way of doing this in go. If this is really such a special case that we need to be extra explicit about copying then maybe. But my concern is that fighting the language like this leads to a mess/a burden we cannot hope to (would not want to) live up to everywhere.

I particularly don't like this given that we are running it through go-wire for little reason.

genesis/maker.go
benjaminbollen: Im ok with introducing the constants and moving them later

genesis/types.go
benjaminbollen: Im not sure whether it is fighting the language or fighting bad habits

benjaminbollen: using go-wire was here rather elegant; otherwise it requires many lines of reflecting which keytype it is then copying it in the correct fixed size byte array and then casting it back to the interface;

in my opinion `interface PubKey` should have a `Clone()` but it does not; so using the available interfaces worked quite nicely

silasdavis: > because it allows you to make it explicit when you're copying data; and as a result avoid copying it when you don't have to.

You are copying data exactly when you don't have to because `PubKey.Bytes()` does a copy itself (as it ought to)

benjaminbollen: but it returns a []byte as type byte prepended public key bytes; I need to then copy that byte slice in a `[32]byte` or `[64]byte` and then cast that to the interface; this one line captures all that;

Im fine for it to be improved in `eris-db/crypto`; but right now I need a way to take deep ownership of the data in genesisDoc so that once the `GenesisDoc` is created it is independent of the memory that went into creating it

silasdavis: My point is you have a PubKey, you don't need a clone, it's the same PubKey `genesisValidator.PubKey` and the `PubKey` interface is immutable.

silasdavis: And you already own the memory because: https://github.com/benjaminbollen/eris-db/blob/441c1a6db5b3d679b94df5bd0335ee32f5d9697e/genesis/maker.go#L77

benjaminbollen:
continued in #506

PR #502, opened 2017-02-21
Hacked away until buildable and runnable with Tendermint 0.8.0

silasdavis:
This gets feature-tendermint_0.8.0 to a state where:

- Tendermint and its transitive dependencies are upgraded 0.8.0
- All reference to tmsp have been converted to abci references
- Our partial support for out-of-process tmsp has been removed (we can add back in the abci server stuff if we want)
- EventSwitch has been changed to an interface throughout
- A Tendermint node can start but fails on making the `cs_wal_dir`

feature-tendermint_0.8.0 is a WIP branch with failing tests. So this is potentially mergeable despite failing tests. We may be able to find a way to divide work at this point, or it may be easier for me to try and get a bit further.


benjaminbollen:
I will merge release-0.16 into eris-ltd:feature-tendermint_0.8.0; so that this PR gives a clearer picture of what has been done and where we are

silasdavis:
have rebased against eris-ltd/feature-tendermint_0.8.0

rpc/tendermint/test/common_test.go
```diff
@@ -1,4 +1,3 @@
-// +build integration
 
 // Space above here matters
```
benjaminbollen: best to remember to remove //space above here matters now ;)

rpc/tendermint/test/config.go
```diff
@@ -135,21 +135,19 @@ genesis_file = "genesis.json"
...
 ################################################################################
 
-[tmsp]
+[abci]
 # listener address for accepting tendermint socket protocol connections
 listener = "tcp://0.0.0.0:46658"
 
 ################################################################################
-##
+##yeah we had partial support for that with TMSP
```
benjaminbollen: we can remove this from the config

PR #507, opened 2017-02-23
Use sha3 has for snative addresses for future-proofing

silasdavis:


manager/eris-mint/evm/snative.go
```diff
@@ -247,14 +247,27 @@ func (contract *SNativeContractDescription) Dispatch(appState AppState,
...
 
-// We define the address of an SNative contact as the simplest possible hash of
-// its canonical name
-func (contract *SNativeContractDescription) Address() Word256 {
-	return LeftPadWord256([]byte(contract.Name))
+// We define the address of an SNative contact as the first 20 bytes of the sha3
+// hash of its name
+func (contract *SNativeContractDescription) Address() abi.Address {
+	var address abi.Address
+	copy(address[:], sha3.Sha3([]byte(contract.Name))[:abi.AddressLength])
```
benjaminbollen: Ethereum takes the right most 20 bytes of the 256bit Keccak digest; not the first 20 bytes;

see eris-keys here https://github.com/eris-ltd/eris-keys/blob/master/crypto/key.go#L152-L153
or also here http://bitcoin.stackexchange.com/questions/42055/what-is-the-approach-to-calculate-an-ethereum-address-from-a-256-bit-private-key

benjaminbollen:
Overall nice improvement

manager/eris-mint/evm/snative.go
silasdavis: that's of a public key though, so it's kind of unrelated... We could copy, but I can't see how it will make any difference. The point is just to have a uniform hash over the 20 bytes.

benjaminbollen:
Yes, I reviewed it last thing before I left, but forgot to merge. lgtm

PR #508, opened 2017-02-23
Fix up consensus/tendermint config to match tendermint 0.8.0 

silasdavis:
This allows the node to run. I think most rpc_client_tests are passing too, with the exception of the blockchain hash test.

### DIFF
--- a/consensus/tendermint/config.go
+++ b/consensus/tendermint/config.go
@@ -51,6 +51,11 @@ func GetTendermintConfig(loadedConfig *viper.Viper) *TendermintConfig {
 //------------------------------------------------------------------------------
 // Tendermint defaults
 
+//
+// Contract
+//
+
+
 func (tmintConfig *TendermintConfig) AssertTendermintDefaults(chainId, workDir,
 	dataDir, rootDir string) {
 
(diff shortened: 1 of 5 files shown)


consensus/tendermint/config.go
```diff
@@ -64,17 +69,21 @@ func (tmintConfig *TendermintConfig) AssertTendermintDefaults(chainId, workDir,
...
 	tmintConfig.SetDefault("db_dir", dataDir)
 	tmintConfig.SetDefault("log_level", "info")
 	tmintConfig.SetDefault("rpc_laddr", "")
 	tmintConfig.SetDefault("prof_laddr", "")
 	tmintConfig.SetDefault("revision_file", path.Join(workDir, "revision"))
-	tmintConfig.SetDefault("cswal", path.Join(dataDir, "cswal"))
-	tmintConfig.SetDefault("cswal_light", false)
+	tmintConfig.SetDefault("cs_wal_dir", path.Join(dataDir,"cs.wal"))
+	tmintConfig.SetDefault("cs_wal_light", false)
+	tmintConfig.SetDefault("filter_peers", false)
```
benjaminbollen: these are great flags to have

consensus/tendermint/config.go
```diff
@@ -64,17 +69,21 @@ func (tmintConfig *TendermintConfig) AssertTendermintDefaults(chainId, workDir,
...
 
-	tmintConfig.SetDefault("block_size", 10000)
+	tmintConfig.SetDefault("block_size", 10000)      // max number of txs
+	tmintConfig.SetDefault("block_part_size", 65536) // part size 64K
```
benjaminbollen: nice!

PR #509, opened 2017-02-24
rpc/tendermint: fix TestBlockchainInfo

benjaminbollen:
rpc/tendermint: fix TestBlockchainInfo as block metadata has restructured on tendermint 0.8


silasdavis:
Yay

PR #520, opened 2017-02-28
Test for json_rpc_server BroadcastTx and refactor

silasdavis:
I was putting this together but unaware of: https://github.com/eris-ltd/eris-db/pull/516

Luckily although the changes did conflict they are kind of complementary. I've avoided extending the `Codec` interface to get the same effect. I've also move things around to more closely our naming elsewhere, or just name things as what they actually are.

mock_pipe is now used by the `json_server_test`


benjaminbollen:
I tried to rebase it to develop; but the PR does not re-calculate for me; despite develop now updated with latest of v0.16

rpc/v0/json_rpc.go
```diff
@@ -12,16 +12,17 @@
...
 import (
 	"encoding/json"
+	"io"
 )
 
 // JSON-RPC 2.0 error codes.
 const (
-	PARSE_ERROR      = -32700
 	INVALID_REQUEST  = -32600
+	PARSE_ERROR      = -32700
```
silasdavis: yeah that wasn't meant to happen

rpc/codec.go
```diff
@@ -1,28 +0,0 @@
-// Copyright 2017 Monax Industries Limited
-//
```
silasdavis: It's not general in the sense of being used by both RPCs. I don't that we'd particularly likely to use it for the next iteration of the RPC

silasdavis:
That's odd, I can reopen against develop.

When you say the rpc interface, you mean what is currently `v0/json_rpc.go`?

Perhaps that can/should go back up. I'll take a look at reopen.

benjaminbollen:
Yes, if you can just take the unit tests and add them that would be great.  Otherwise let me know and I can do that; you should enjoy your holidays

silasdavis:
I've merge in my changes with develop (at this stage cleaner than a rebase). 

I've pulled the jsonrpc.go potentially usable in a v1 up to the root of rpc and just called it 'rpc.go' for generic definitions as you suggest.

I still think the renames and other changes are a win, so I've included them too, not just the tests. They are more reflective of what things are actually doing and they adhere to our snake case and full names conventions.

PR #525, opened 2017-03-04
strings.ToUpper(the-license-filename-prefix)

zramsay:
- the GitHub feature that displays the license type in repo header seems to require `LICENSE.md` and not `license.md` & it's also the default for a license filename; there's no reason to deviate.

cli (has it): 
screenshot from 2017-03-03 21-32-02

edb (doesn't have it): 
screenshot from 2017-03-03 21-32-30


benjaminbollen:
LGTM

benjaminbollen:
the only pain is that merging this will change the hash of the docker image. Days without immutability: set back to zero.

PR #531, opened 2017-03-21
docs: update harmonized links; remove generator

zramsay:
- removes redundant generator files for docs generation
- updates links throughout to reflect new documentation paradigm
- merge only when we go live with the new docs on the website (shortly after release)


cmd/eris-db.go
```diff
@@ -34,7 +34,7 @@ your needs.
 
 Made with <3 by Monax Industries.
 
-Complete documentation is available at https://monax.io/docs/documentation
+Complete documentation is available at https://monax.io/docs
```
benjaminbollen: eris-db will need its own documentation (especially with Hyperledger now); but yes, good change

benjaminbollen:
CircleCI is broken under renaming of organisation; but PR concerns documentation, so will merge without CI result.

PR #532, opened 2017-03-21
Just vendor in the dependencies

VoR0220:
Reasons:

1. It makes our docker images smaller

2. It makes tooling's life a lot easier in importing dependencies.

3. It allows us to drop tools such as hell

I think these 3 succint reasons are more than enough to get rid of this paradigm.


benjaminbollen:
This should definitely not be a PR to master

benjaminbollen:
For now closing PR, as it is too uncomfortable to have one click-away from merging.  We can have the discussion further as already had and restarted.

VoR0220:
Reopening PR to bring back discussion of vendoring in dependencies

Dockerfile
```diff
@@ -5,8 +5,6 @@ MAINTAINER Monax <support@monax.io>
 ENV TARGET eris-db
 ENV REPO $GOPATH/src/github.com/monax/$TARGET
 
-ADD ./glide.yaml $REPO/
-ADD ./glide.lock $REPO/
 WORKDIR $REPO
 RUN glide install
```
benjaminbollen: if you re vendoring in the dependencies then running glide install is not necessary anymore either.

VoR0220: Agreed. Let me kill that.

silasdavis:
We don't need this PR to have the discussion, we're already planning to do something in this direction fairly soon.

PR #544, opened 2017-03-29
[Hotfix] Use debug flag

silasdavis:
Noticed the debug flag was not being respected. The EVM debug ought to be handled by the logging system, but this is still pending...


cmd/serve.go
```diff
@@ -161,6 +162,8 @@ func ServeRunner(do *definitions.Do) func(*cobra.Command, []string) {
 				DefaultConfigFilename)
 		}
 
+		vm.SetDebug(do.Debug)
```
benjaminbollen: ugh. that is a very direct way to pull this through the whole stack... but  you might be right that for now it is a useful things to have.

PR #553, opened 2017-04-04
Rename to Burrow...

silasdavis:
This omits some anticipated upstream changes to the location of our docker images on quay.io so another PR will be needed before we reach a commit we can actually release.


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/eris/build
+FROM quay.io/monax/build
```
zramsay: can be versioned: https://github.com/monax/cli/pull/1295/files#diff-0551cf25c62a749fe9b04d036a1ef478R44

Dockerfile.deploy
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/monax/base:alpine
+FROM quay.io/monax/base
```
zramsay: version it? that's what i did for keys and compilers

silasdavis: uh yes... need to do that

PR #565, opened 2017-04-11
Release 0.16

silasdavis:
Merge for tagging release on master


config/templates.go
```diff
@@ -280,7 +280,7 @@ const sectionBurrowMint = `
 ################################################################################
 ##
 ## Burrow-Mint
-## version 0.16
+## version 0.17
```
zramsay: this is 0.17

rpc/tendermint/test/config.go
```diff
@@ -38,7 +39,7 @@ var defaultConfig = `# Copyright 2017 Monax Industries Limited
 assert_chain_id = "MyChainId"
 # semantic major and minor version
 major_version = 0
-minor_version = 12
+minor_version = 16
```
zramsay: this is 0.16

PR #568, opened 2017-04-12
pull in vendor

VoR0220:
Signed-off-by: RJ Catalano <rj@monax.io>


silasdavis:
We are discussing how to handle vendor, we don't need a PR for this

PR #573, opened 2017-04-13
Pull in permissions util fix

silasdavis:
Needed for 0.16.2


benjaminbollen:
LGTM (effectlively develop v0.17 is on par now with v0.16.2 modulo version number)

PR #574, opened 2017-04-13
[develop] Apply hotfix from PR #571

silasdavis:


benjaminbollen:
LGTM (we're running around a bit but we should have landed now :) )

PR #576, opened 2017-04-13
Upgrade CI to docker 1.10.0 to fix remove image issue

silasdavis:
This PR:

- Upgrades to Circle approved docker 1.10.0
- Fixes issue where build image could not be removed
- Reintroduces removal of build image so it is not pushed
- Stops using repo username and instead hard codes hyperledger/burrow so CI will work against forks

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,21 +1,21 @@
 machine:
   environment:
     GOPATH: $HOME/.go_workspace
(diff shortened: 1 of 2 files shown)


silasdavis:
@benjaminbollen this is ready to reviewed/merged. I'm fairly confident it will restore our docker pushing to how it was before the move to hyperledger.

PR #593, opened 2017-04-29
Tendermint 0.9.2 upgrade and rework versioning 

silasdavis:
This PR:
- Makes changes necessary to support Tendermint 0.9.2 and upgrades our dependencies to match
- Adds attempt to gracefully shutdown `ConsensusEngine`
- Updates RPC tests to use genesis making and config generation rather than hard-coded values

In the process of doing the upgrade I came up against the slightly cumbersome process for updating our versioning. Since the upgrade touched this I have tried to address it here. I understand the legitimate motivations for the independent versioning of modules (from Burrow itself), and I am happy to rollback changes if this really goes too far. However I put forward the following arguments/principles that guided my work:

1. Burrow should continue to support the idea of 'modules' (consensus and manager), however it is down to a point release (that is specific compiled binary) of Burrow to ensure that the modules it bundles are compatible (with respect to versions). We do this through dependency management and versioning and therefore it is redundant to further assert on the versions of our modules. In reality it is not clear at all how we could support multiple versions of 'the same' manager. If we would like to support different compatibility-level versions of a manager or consensus engine then it is sufficient to give them separate names. For the purposes of Burrow's initialisation it does not need to know they they belong to the same lineage.

2. To determine compatibility between modules we should try and make use of the type system rather than hand-coded values (that we have to robotically have to upgrade in multiple places). We can extend the current 'dependency injection' of `ConsensusEngine` to help us with this. When a consensus module is loaded into a 'manager `Pipe`' (we load `Manager` first so we may as well consider the pipe to be manager-typed at that point) we can ask `ConsensusEngine` and `Manager`/`Application` to check `Pipe` for compatible implementations by trying to type convert interfaces: `_, ok := pipe.ConsensusEngine.(BurrowMintCompatibleConsensusEngine)`. We can capture the actual functional requirements that a module may have by using this pattern. If we really need to be version specific (i.e. we have compatible interface signatures that belonging to logically incompatible versions) we can add a marker function `func BurrowMint_0_18_4_Consensus()` that captures that incompatible, but we can do so when we really need it rather than pre-emptively.

3. It is useful to error on being provided a configuration written by a different version of Burrow to the runtime version of Burrow. To do this we can continue to close the Burrow version into the generated config file. But as in 1 it is sufficient to just use the single over-arching version number.

The last commit introduces mechanisms to cover points 2 and 3


silasdavis:
Closing in favour of split:

https://github.com/hyperledger/burrow/pull/595
https://github.com/hyperledger/burrow/pull/596

Issue #583, opened 2017-04-20
Update Tendermint to v0.9


benjaminbollen:
fixes included in tendermint v0.9:
- handshake issue on restarting state https://github.com/tendermint/tendermint/issues/388

PR #595, opened 2017-05-02
Upgrade to Tendermint 0.9.2

silasdavis:
This PR:

- Makes changes necessary to support Tendermint 0.9.2 and upgrades our dependencies to match
- Adds attempt to gracefully shutdown ConsensusEngine
- Updates RPC tests to use genesis making and config generation rather than hard-coded values


core/core.go
```diff
@@ -119,3 +119,8 @@ func (core *Core) NewGatewayTendermint(config *server.ServerConfig) (
 	return rpc_tendermint.NewTendermintWebsocketServer(config,
 		core.tendermintPipe, core.evsw)
 }
+
+// Stop the core allowing for a graceful shutdown of component in order.
+func (core *Core) Stop() bool {
```
benjaminbollen: Nice

manager/burrow-mint/burrow-mint.go
```diff
@@ -177,6 +186,32 @@ func (app *BurrowMint) Commit() (res abci.Result) {
...
+// Signals the beginning of a block
+func (app *BurrowMint) BeginBlock(hash []byte, header *abci.Header) {
+
 }
+
+// Signals the end of a blockchain, return value can be used to modify validator
+// set and voting power distribution see our BlockchainAware interface
+func (app *BurrowMint) EndBlock(height uint64) (respEndblock abci.ResponseEndBlock) {
+	// TODO: [Silas] Bondage
+	// TODO: [Silas] this might be a better place for us to dispatch new block
```
benjaminbollen: Agreed.

silasdavis: Actually only issue with that is is that I think they are sending ABCI EndBlock _before_ they save Tendermint state. Not an issue for something things, but may be if you were expecting post-block consensus state to be ready.

benjaminbollen:
closes https://github.com/hyperledger/burrow/issues/583

PR #598, opened 2017-05-04
Fix broken unsubscribe…

### LINKED ISSUE #597: race condition in tests on NewBlock event
Race condition in test on new block event for tendermint v0.9.2 on (develop / 0.17)

```
github.com/hyperledger/burrow/rpc/tendermint/test
=== RUN   TestSubscribe
--- FAIL: TestSubscribe (1.01s)
    websocket_client_test.go:298: Continued to see NewBlock event after unsubscribing
```

- `burrow version` 0.17.0

silasdavis:
Also don't rely on go-rpc's definition of `Unsubscribe` and `Subscribe`, which even when it was compatible with ours was only 'by chance' (and history). So we:

- Use 'eventId' arg name for `Subscribe` consistently
- Use 'subscriptionId' arg name for `Unsubscribe` consistently
- Centralise subscribe/unsubsribe in `rpc/tendermint/client` package (used by tests and `NodeClient`) in a similar way as the non-websocket call-based methods are.

fixes #597

### DIFF
--- a/rpc/tendermint/test/websocket_client_test.go
+++ b/rpc/tendermint/test/websocket_client_test.go
@@ -260,11 +260,12 @@ func TestSubscribe(t *testing.T) {
 	var subId string
 	subscribe(t, wsc, txs.EventStringNewBlock())
 
-	timeout := time.NewTimer(timeoutSeconds * time.Second)
+	// timeout to check subscription process is live
+	timeout := time.After(timeoutSeconds * time.Second)
 Subscribe:
 	for {
 		select {
-		case <-timeout.C:
+		case <-timeout:
 			t.Fatal("Timed out waiting for subscription result")
 
 		case bs := <-wsc.ResultsCh:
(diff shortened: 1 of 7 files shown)


rpc/tendermint/test/websocket_client_test.go
```diff
@@ -260,29 +260,32 @@ func TestSubscribe(t *testing.T) {
...
-		case <-timeout.C:
+		case <-timeout:
 			t.Fatal("Timed out waiting for subscription result")
 
 		case bs := <-wsc.ResultsCh:
 			resultSubscribe, ok := readResult(t, bs).(*core_types.ResultSubscribe)
 			if ok {
 				assert.Equal(t, txs.EventStringNewBlock(), resultSubscribe.Event)
 				subId = resultSubscribe.SubscriptionId
+				fmt.Printf("\x1b[30m\x1b[46mMARMOT!: subid %s\x1b[0m\n", subId)
```
benjaminbollen: Battle marmot needs to return to base

silasdavis: Haha. Dammit those special ops marmots can be hell to extract from the field once they've gone to ground.

PR #605, opened 2017-05-05
Clean up storage and account tombstones from BlockCache 

### LINKED ISSUE #542: Call to selfdestruct(..) stops the chain
Please include in your bug report:

- [ ] `eris-db version` (docker image tag or branch if built from source)
0.12.0
- [ ] `eris version` (if applicable)
0.12.0
- [ ] `go version` (if applicable)
- [ ] `docker version` (if applicable)
1.13.0
- [ ] operating system details (osx/windows/linux)
osx
- [ ] what you expected to happen
chain keeps alive
- [ ] what actually happened
chain stops
- [ ] steps to reproduction
I'm making a call to Solidity selfdestruct via a cascading contract call, e.g.:

From a TaskManager contract:
```
    function removeTask(address manager, address taskAddress)
    returns (bool result) {
        Task t = Task(taskAddress);
        if (t.remove(manager)) {
            taskList.removeElement(t);
            TaskDeleted(t, "Task Deleted", block.number, block.timestamp);
            return true;
        } else {
            return false;
        }
    }
```
call from within the Task contract:

```
    function remove(address manager)
    onlyManager(manager)
    isEmpty(manager)
    returns (bool)
    {
        selfdestruct(owner);
        return true;
    }
```
It doesn't matter if I use the `owner` or the `manager` address for the refund

silasdavis:
An issue occurs on block commit when  an account is removed from the `BlockCache`, for example (the only example I'm aware of) when `SELFDESTRUCT`/`SUICIDE` is called. This line in `BlockCache`.Sync is reached: https://github.com/hyperledger/burrow/blob/master/manager/burrow-mint/state/block_cache.go#L227 when Burrow attempts to remove an account from the merkle IAVL tree when it has already been removed. This happens one block after the block that _did_ successfully remove the account because the `BlockCache` holds an 'append only' cache of account and storage changes: https://github.com/hyperledger/burrow/blob/master/manager/burrow-mint/state/block_cache.go#L41-L42. It stores an account removal by setting a 'removed' flag so the account object becomes a tombstone object. But the `Sync` method does not clean up these objects so on the subsequent block it attempts to remove an account that has already been removed.

I'd like to think a bit more about where else we may be having a stale cache and see if there is a reasonable way to test the Sync method if that can be added in reasonably timely fashion.

fixes #542 

Resolves my test case that can be found here: https://gist.github.com/silasdavis/7f35cb31a255c18d2fc70eed1fa38211

### DIFF
--- a/manager/burrow-mint/state/block_cache.go
+++ b/manager/burrow-mint/state/block_cache.go
@@ -197,6 +197,9 @@ func (cache *BlockCache) Sync() {
 			curStorage = storage
 		}
 		if curAccRemoved {
+			// We should delete any storage for removed accounts in anticipation of
+			// those accounts being reaped from the cache
+			delete(cache.storages, storageKey)
 			continue
 		}
 		value, dirty := cache.storages[storageKey].unpack()
@@ -224,8 +227,12 @@ func (cache *BlockCache) Sync() {
 		if removed {
 			removed := cache.backend.RemoveAccount([]byte(addrStr))
 			if !removed {
-				sanity.PanicCrisis(fmt.Sprintf("Could not remove account to be removed: %X", acc.Address))
+				sanity.PanicCrisis(fmt.Sprintf("Could not remove account to be removed: %X", addrStr))
 			}
+			// Since attempting to delete an already deleted account leads to a
+			// PanicCrisis (above) we must make sure we remove the tombstones of
+			// removed accounts
+			delete(cache.accounts, addrStr)
 		} else {
 			if acc == nil {
 				continue
@@ -259,6 +266,8 @@ func (cache *BlockCache) Sync() {
 			if !removed {
 				sanity.PanicCrisis(fmt.Sprintf("Could not remove namereg entry to be removed: %s", nameStr))
 			}
+			// Remove the tombstone name reg object from the cache
+			delete(cache.names, nameStr)
 		} else {
 			if entry == nil {
 				continue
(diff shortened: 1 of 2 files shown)


manager/burrow-mint/state/block_cache.go
```diff
@@ -224,8 +225,9 @@ func (cache *BlockCache) Sync() {
 		if removed {
 			removed := cache.backend.RemoveAccount([]byte(addrStr))
 			if !removed {
-				sanity.PanicCrisis(fmt.Sprintf("Could not remove account to be removed: %X", acc.Address))
+				sanity.PanicCrisis(fmt.Sprintf("Could not remove account to be removed: %X", addrStr))
```
silasdavis: If we do get here better to not get a nil pointer error before the panic sanity if acc == nil

silasdavis:
Although this fixes the current issues with out-of-sync cache. I am questioning whether an attempt to remove a removed or non-existent account (or name reg entry) should ever cause a panic. If we think it should (what does this save us from?) then there is still potential for a panic in the present code if we add and remove an account (or name reg entry) without `BlockCache.Sync()`ing inbetween. So if panics should remain, we should probably fix that now too.

benjaminbollen:
LGTM; but this is not a transparant matter, so I will hold off the merge and come back to it in the morning.

silasdavis:
Also see my comment above. That needs a resolution before any merge.

benjaminbollen:
I would advice that we merge #605 with the current fix; because if our understanding is correct then the reported failures on selfdestruct should be addressed (assuming no-one instantly destructs new accounts); then we raise the issue of the removal or improved knowledge for required panic in a new item.

benjaminbollen:
LGTM

PR #606, opened 2017-05-06
Change hasBase to return bool

silasdavis:
This change was inadvertently omitted from #500, it changes the return type of `hasBase` to bool to match the other predicate methods.

I've also added an error message if one attempts to call an snative directly (simulated call as in query-contract) which mirrors the error message given in execution.go when doing the same via a `CallTx`.

### DIFF
--- a/manager/burrow-mint/evm/native.go
+++ b/manager/burrow-mint/evm/native.go
@@ -24,8 +24,8 @@ import (
 
 var registeredNativeContracts = make(map[Word256]NativeContract)
 
-func RegisteredNativeContract(addr Word256) bool {
-	_, ok := registeredNativeContracts[addr]
+func RegisteredNativeContract(address Word256) bool {
+	_, ok := registeredNativeContracts[address]
 	return ok
 }
 
--- a/manager/burrow-mint/pipe.go
+++ b/manager/burrow-mint/pipe.go
@@ -409,6 +409,11 @@ func (pipe *burrowMintPipe) DumpStorage(address []byte) (*rpc_tm_types.ResultDum
 // TODO: [ben] resolve incompatibilities in byte representation for 0.12.0 release
 func (pipe *burrowMintPipe) Call(fromAddress, toAddress, data []byte) (*rpc_tm_types.ResultCall,
 	error) {
+	if vm.RegisteredNativeContract(word256.LeftPadWord256(toAddress)) {
+		return nil, fmt.Errorf("Attempt to call native contract at address "+
+			"%X, but native contracts can not be called directly. Use a deployed "+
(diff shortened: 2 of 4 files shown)


manager/burrow-mint/pipe.go
```diff
@@ -409,6 +411,10 @@ func (pipe *burrowMintPipe) DumpStorage(address []byte) (*rpc_tm_types.ResultDum
 // TODO: [ben] resolve incompatibilities in byte representation for 0.12.0 release
 func (pipe *burrowMintPipe) Call(fromAddress, toAddress, data []byte) (*rpc_tm_types.ResultCall,
 	error) {
+	if vm.RegisteredNativeContractAddress(toAddress) {
+		return nil, errors.New("Native contracts can not be called directly." +
+			"Use a deployed contract that calls the s")
```
benjaminbollen: unfinished error message, include `toAddress`, spacing between sentences and that native contracts cannot be called directly "... from RPC/pipe".

please apply `LeftPadWord256` here explicitly.

Even though this error is caught before entering execution, and additionally it would hit L420 `GetAccount` and should fail; but it is a good catch because we should consider blocking out the (creation and) existence of contracts at a registered native contracts)

manager/burrow-mint/evm/native.go
```diff
@@ -24,8 +24,12 @@ import (
 
 var registeredNativeContracts = make(map[Word256]NativeContract)
 
-func RegisteredNativeContract(addr Word256) bool {
-	_, ok := registeredNativeContracts[addr]
+func RegisteredNativeContractAddress(address []byte) bool {
```
benjaminbollen: We should work towards stronger typing, not looser typing;  `LeftPadWord256` should be called by caller of `RegisteredNativeContract(word256) bool`; so please revert and adjust two places where you call.

manager/burrow-mint/state/execution.go
```diff
@@ -430,8 +430,10 @@ func ExecTx(blockCache *BlockCache, tx txs.Tx, runCall bool, evc events.Fireable
 				return txs.ErrTxInvalidAddress
 			}
 			// check if its a native contract
-			if vm.RegisteredNativeContract(LeftPadWord256(tx.Address)) {
-				return fmt.Errorf("NativeContracts can not be called using CallTx. Use a contract or the appropriate tx type (eg. PermissionsTx, NameTx)")
+			if vm.RegisteredNativeContractAddress(tx.Address) {
```
benjaminbollen: please revert `LeftPadWord256`; such conversion from []byte to Address should be handled by a type `Address`, not by a function signature.  Until we introduce such stronger typing the paradigm is to use these conversion functions.

PR #607, opened 2017-05-09
Implement dynamic memory and fix out of bounds issues

### LINKED ISSUE #258: Unable to return array from contract
Hi,
I have a simple contract which returns few arrays. Works fine in the online solidity compiler. But after deploying it to the chain and invoking it from the app.js its giving the following error. 

Have found a bug related to this https://github.com/eris-ltd/eris-db.js/issues/27

Any particular reason why this error is occurring?  

My contract is like: 
```javascript
function getMarket() constant returns(uint[], uint[], uint[], uint[]) {
    uint[] memory optionIDs = new uint[](20);
    uint[] memory strikes_ = new uint[](20);
    uint[] memory positions_ = new uint[](20);
    uint[] memory cashes = new uint[](20);
    uint z = 0;
      for (uint i=0; i<10; i++) {
        optionIDs[i] = i;
        strikes_[i] = i+10;
        positions_[i] = i+100;
        cashes[i] = i+50;
        z++;
      }
    return (optionIDs, strikes_, positions_, cashes);
  }
```

Error is:
```
[JSON RPC 2.0 Error] Message Id: 1, (INTERNAL_ERROR) - Error when transacting: Memory out of bounds
[Error: Error when transacting: Memory out of bounds]
Error callback from sendTransaction

/home/azureuser/.eris/apps/test/app.js:50
    if (error) { throw error; }
                       ^
Error: Error when transacting: Memory out of bounds
    at /home/azureuser/.eris/apps/test/node_modules/eris-contracts/node_modules/eris-db/lib/rpc/http.js:79:22
    at httpRequest.onreadystatechange (/home/azureuser/.eris/apps/test/node_modules/eris-contracts/node_modules/eris-db/lib/rpc/http.js:103:17)
    at dispatchEvent (/home/azureuser/.eris/apps/test/node_modules/eris-contracts/node_modules/eris-db/node_modules/xmlhttprequest/lib/XMLHttpRequest.js:572:25)
    at setState (/home/azureuser/.eris/apps/test/node_modules/eris-contracts/node_modules/eris-db/node_modules/xmlhttprequest/lib/XMLHttpRequest.js:591:14)
    at IncomingMessage.<anonymous> (/home/azureuser/.eris/apps/test/node_modules/eris-contracts/node_modules/eris-db/node_modules/xmlhttprequest/lib/XMLHttpRequest.js:431:13)
    at IncomingMessage.EventEmitter.emit (events.js:117:20)
    at _stream_readable.js:920:16
    at process._tickCallback (node.js:415:13)
```

### LINKED ISSUE #300: [EVM] String Concatenation fails (invalid Jump Dest / Memory out of bounds)
See https://github.com/eris-ltd/eris-db.js/issues/27 for details.

I've recreated the bug here (https://github.com/eris-ltd/eris-contracts.js/blob/beta/test/integration/failing/%2327.js):

``` javascript

'use strict'

const assert = require('assert')
const Promise = require('bluebird')
const test = require('../../lib/test')

const source = `
  contract twentyseven {
    function getString2() constant returns (string){
        string memory abcde = new string(3);
        return "a";
    }
  }
`

it('tests issue #27', function () {
  this.timeout(60 * 1000)

  return test.newContractManager('blockchain').then((contractManager) => {
    const compiled = test.compile(source).twentyseven
    const abi = JSON.parse(compiled.interface)
    const bytecode = compiled.bytecode
    const contractFactory = contractManager.newContractFactory(abi)

    return Promise.fromCallback((callback) =>
      contractFactory.new({data: bytecode}, callback)
    ).then((contract) =>
      Promise.fromCallback((callback) => contract.getString2(callback))
        .then((result) => {
          assert.equal(result, 'a')
        })
    )
  })
})
```

### LINKED ISSUE #474: Valid solidity function leads to "Invalid jump dest" in ErisDB
The following Test contract runs fine in browser-solidity using both Solidity compiler 0.3.6 and 0.4.9 (latest). However, executing this code on ErisDB leads to an error:

```
Executing Job                                 testBytes32ToString
Encountered Exception from chain: Transaction confirmed with exception:%!(EXTRA string=Invalid jump dest)
```

```
contract Test {
       function toString (bytes32 x) constant returns (string) {
        bytes memory bytesString = new bytes(32);
        uint charCount = 0;
        for (uint j = 0; j < 32; j++) {
            byte char = byte(bytes32(uint(x) * 2 ** (8 * j)));
            if (char != 0) {
                bytesString[charCount] = char;
                charCount++;
            }
        }
        bytes memory resultBytes = new bytes(charCount);
        for (j = 0; j < charCount; j++) {
            resultBytes[j] = bytesString[j];
        }

        return string(resultBytes);
    }
}
```

Versions:
OSX version: 10.11.6
Eris CLI Version: 0.12.0
Docker:
Client:
 Version:      1.13.0
 API version:  1.25
 Go version:   go1.7.3
 Git commit:   49bf474
 Built:        Wed Jan 18 16:20:26 2017
 OS/Arch:      darwin/amd64
Server:
 Version:      1.13.0
 API version:  1.25 (minimum version 1.12)
 Go version:   go1.7.3
 Git commit:   49bf474
 Built:        Wed Jan 18 16:20:26 2017
 OS/Arch:      linux/amd64
 Experimental: false
docker-machine version 0.9.0, build 15fd4c7

silasdavis:
This PR introduces a dynamic memory implementation that satisfies the assumption (by Solidity) that memory will be dynamically allocated at the memory boundary when an `MSTORE` writes to the memory boundary provided by `MSIZE`.

The Solidity compiler assumes that memory will be dynamically allocated at the current memory upper bounds (as reported by `MSIZE`) if an `MSTORE` operation tries to write to it. I believe the reason for this behaviour is that it is an easy way to guarantee that a portion a of memory will be unused (since it has not yet been allocated). Obviously this only works if the EVM environment will dynamically allocate that memory for you. This PR implements this in Burrow.

For reference, the Solidity assumption around object creation is encoded in the Solidity compiler here: https://github.com/ethereum/solidity/blob/develop/libsolidity/codegen/ExpressionCompiler.cpp#L842-L851

And you can see the dynamic memory allocation in go-ethereum here: https://github.com/ethereum/go-ethereum/blob/master/core/vm/interpreter.go#L152-L173 . Where the memory size is just the mstore offset + one word calculated here: https://github.com/ethereum/go-ethereum/blob/master/core/vm/memory_table.go#L49-L51.

I have taken a simpler approach of growing the memory on a `Write` operation up to a configurable bounds. I make use of the underlying slice implementation in Go to efficiently grow arrays. This PR includes:

- Simple `Memory` interface to encapsulate memory operations
- Adds a `memoryProvider` argument to `NewVM`
- Switches out memory slice for new `Memory` object 
- Fixes issues arising because the `subslice` function simply wasn't fit for purpose
- Adds a variety of tests, including on in the VM to test for the specific 'allocation on boundary' behaviour Solidity expects.

I have verified that this fixes the following issues:

fixes #474 
fixes #258 
fixes #300 (could not use exactly the same code)

### DIFF
--- a/manager/burrow-mint/evm/snative.go
+++ b/manager/burrow-mint/evm/snative.go
@@ -87,10 +87,10 @@ func SNativeContracts() map[string]*SNativeContractDescription {
 			`,
 				"addRole",
 				[]abi.Arg{
-					arg("_account", abi.AddressTypeName),
-					arg("_role", roleTypeName),
+					abiArg("_account", abi.AddressTypeName),
+					abiArg("_role", roleTypeName),
 				},
-				ret("result", abi.BoolTypeName),
+				abiReturn("result", abi.BoolTypeName),
 				ptypes.AddRole,
 				addRole},
 
@@ -102,10 +102,10 @@ func SNativeContracts() map[string]*SNativeContractDescription {
 			`,
 				"removeRole",
 				[]abi.Arg{
-					arg("_account", abi.AddressTypeName),
-					arg("_role", roleTypeName),
+					abiArg("_account", abi.AddressTypeName),
+					abiArg("_role", roleTypeName),
 				},
-				ret("result", abi.BoolTypeName),
+				abiReturn("result", abi.BoolTypeName),
 				ptypes.RmRole,
 				removeRole},
 
@@ -117,10 +117,10 @@ func SNativeContracts() map[string]*SNativeContractDescription {
 			`,
 				"hasRole",
 				[]abi.Arg{
-					arg("_account", abi.AddressTypeName),
-					arg("_role", roleTypeName),
+					abiArg("_account", abi.AddressTypeName),
+					abiArg("_role", roleTypeName),
 				},
-				ret("result", abi.BoolTypeName),
+				abiReturn("result", abi.BoolTypeName),
 				ptypes.HasRole,
 				hasRole},
 
@@ -133,11 +133,11 @@ func SNativeContracts() map[string]*SNativeContractDescription {
 			`,
 				"setBase",
 				[]abi.Arg{
-					arg("_account", abi.AddressTypeName),
-					arg("_permission", permFlagTypeName),
-					arg("_set", abi.BoolTypeName),
+					abiArg("_account", abi.AddressTypeName),
+					abiArg("_permission", permFlagTypeName),
+					abiArg("_set", abi.BoolTypeName),
 				},
-				ret("result", permFlagTypeName),
+				abiReturn("result", permFlagTypeName),
 				ptypes.SetBase,
 				setBase},
 
@@ -149,9 +149,9 @@ func SNativeContracts() map[string]*SNativeContractDescription {
       `,
 				"unsetBase",
 				[]abi.Arg{
-					arg("_account", abi.AddressTypeName),
-					arg("_permission", permFlagTypeName)},
-				ret("result", permFlagTypeName),
+					abiArg("_account", abi.AddressTypeName),
+					abiArg("_permission", permFlagTypeName)},
+				abiReturn("result", permFlagTypeName),
 				ptypes.UnsetBase,
 				unsetBase},
 
@@ -163,9 +163,9 @@ func SNativeContracts() map[string]*SNativeContractDescription {
 			`,
 				"hasBase",
 				[]abi.Arg{
-					arg("_account", abi.AddressTypeName),
-					arg("_permission", permFlagTypeName)},
-				ret("result", abi.BoolTypeName),
+					abiArg("_account", abi.AddressTypeName),
+					abiArg("_permission", permFlagTypeName)},
+				abiReturn("result", abi.BoolTypeName),
 				ptypes.HasBase,
 				hasBase},
 
@@ -177,9 +177,9 @@ func SNativeContracts() map[string]*SNativeContractDescription {
 			`,
 				"setGlobal",
 				[]abi.Arg{
-					arg("_permission", permFlagTypeName),
-					arg("_set", abi.BoolTypeName)},
-				ret("result", permFlagTypeName),
+					abiArg("_permission", permFlagTypeName),
+					abiArg("_set", abi.BoolTypeName)},
+				abiReturn("result", permFlagTypeName),
 				ptypes.SetGlobal,
 				setGlobal},
 		),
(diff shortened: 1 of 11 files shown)


manager/burrow-mint/evm/snative.go
```diff
@@ -324,14 +324,14 @@ func (function *SNativeFunctionDescription) NArgs() int {
 	return len(function.Args)
 }
 
-func arg(name string, abiTypeName abi.TypeName) abi.Arg {
+func abiArg(name string, abiTypeName abi.TypeName) abi.Arg {
```
silasdavis: This was renamed to be consistent with ret below

manager/burrow-mint/evm/snative.go
```diff
@@ -324,14 +324,14 @@ func (function *SNativeFunctionDescription) NArgs() int {
...
 	return abi.Arg{
 		Name:     name,
 		TypeName: abiTypeName,
 	}
 }
 
-func ret(name string, abiTypeName abi.TypeName) abi.Return {
+func abiReturn(name string, abiTypeName abi.TypeName) abi.Return {
```
silasdavis: This was renamed so as to not clash with the local variable `ret` in the vm package

benjaminbollen:
please fmt

silasdavis:
I have just amended the commit to debug log the `memErr`s otherwise the same, did make fmt, and tweaked a couple of comments.

1. A thought on this. For `MLOAD` we are hitting an unnecessary copy when we push the :

```go
func LeftPadWord256(bz []byte) (word Word256) {
	copy(word[32-len(bz):], bz)
	return
}
```

For cases like this (where we want a copy _somewhere_ downstream) we could provide a function on the `Memory` interface `func ReadTo(offset, length int64, destination []byte) ([]byte, error)`. Then implement `LeftPadWord256` as `ReadTo(offset, length, word[32-length:])`. The other main occasion that we `Read` is to get the input for calls `CALL`s ultimately these are pushed to stack (obligatory copy). So if we passed `input []byte` from `call` as `input Memory` we could postpone the copy and have `ReadTo` do it whilst still maintaining safety. Again... probably not an optimisation we need now...

2. I'm logging `memErr` at least for now.

3. If we were to sort out our gas pricing to be fit for our purposes then we wouldn't be doing any worse than other famous EVM implementations with this memory model. Essentially memory is priced rather cheaply (1 `MSTORE` cost per 16 MiB!), but at least it is bounded by a linear function in number of `MSTORE`s (alright also `CALLDATACOPY` and others, but still)

silasdavis:
CLI tests pass for me at this version: https://github.com/monax/cli/blob/6a14d0cf2e890be98d9e35f4d26dd29d5292f793/Makefile#L92

benjaminbollen:
Further tests in monax/cli also pass locally for me.

PR #613, opened 2017-05-21
Step back from the notion of two independent logging channels

silasdavis:
When I first implemented the logger I wanted to encourage a high frequency of log messages to the trace channel with impunity. To do this their are two independent language-level channels in InfoTraceLogger. It seemed like a good idea at the time. But in fact it introduces a fair bit of accidental complexity and some odd corner cases (since you want Info and Trace messages to be strictly ordered and interlaced you need to send Info messages to both the Info and Trace output loggers, which then obscures the intent of the sender... it ends up a bit confused).

The desired effect can be achieved in a simpler way with sinks (just filter trace messages), and I will be introducing logging presets (that is preset LoggingConfigs, because it's too much to ask people to write these from scratch for common cases) which will make it even easier.

So putting this down to experience I've decided it was optimising the wrong thing, and too early. So I have:

- Kept the same InfoTraceLogger type definition and semantics (more or less)
- Simplified InfoTraceLogger to take single output logger (which makes InfoTraceLogger a simple wrapper on top of go-kit/log/logger)
- Upgraded to latest go-kit/log which makes a minor improvements (Tendermint are also basing their logging component around go-kit/log)

Non-blocking logging is maintained, which i still thing is a good idea.

Logging presets will follow this, which I hope to be the last 'user friendliness' element that is missing from the logging component. Once we are on Tendermint 0.10.0 we can also provide a better logging capturing experience for Tendermint logs.


logging/lifecycle/lifecycle.go
```diff
@@ -91,14 +89,10 @@ func CaptureStdlibLogOutput(infoTraceLogger types.InfoTraceLogger) {
 }
 
 // Helpers
-func infoTraceLoggersFromLoggingConfig(loggingConfig *config.LoggingConfig) (kitlog.Logger, kitlog.Logger, error) {
-	infoOnlyLogger, _, err := loggingConfig.InfoSink.BuildLogger()
+func infoTraceLoggersFromLoggingConfig(loggingConfig *config.LoggingConfig) (kitlog.Logger, error) {
```
benjaminbollen: name of helper function should be singular now

silasdavis:
Changed that

benjaminbollen:
LGTM

PR #614, opened 2017-05-23
Remove hell

silasdavis:
Since factored out and improved on in https://github.com/silasdavis/glide-lock-transitive


benjaminbollen:
can you then also remove `Masterminds/glide` from `glide.yaml` https://github.com/hyperledger/burrow/blob/master/glide.yaml#L33, as that was introduced together with `hell`

silasdavis:
@benjaminbollen done!

benjaminbollen:
LGTM

PR #621, opened 2017-05-31
apply the megacheck code vet tool

zramsay:
- applies _some_ of the recommendations from the `megacheck` tool, which itself runs `unused`, `gosimple`, and `staticcheck`. See https://github.com/dominikh/go-tools/tree/master/cmd/megacheck for more information.
- only the least controversial edits were made. Bulk removal of unused functions/variables left untouched as they may have a planned usage.
- introduces `make megacheck` to the Makefile with the eventual goal of adding it to the test sequence.


manager/burrow-mint/evm/sha3/keccakf.go
```diff
@@ -40,7 +40,7 @@ var rc = [...]uint64{
 // ro_xx represent the rotation offsets for use in the χ step.
 // Defining them as const instead of in an array allows the compiler to insert constant shifts.
 const (
-	ro_00 = 0
+	ro_00 = 0 // not used
```
silasdavis: presume you did this?

zramsay: yeah, because the `unused` tool throws. can remove

silasdavis:
LGTM

PR #624, opened 2017-07-10
Develop: fixed Tx&Hold bug, added Broadcast&Hold

ratranqu:
1/ Fix TransactAndHold bug when the the contract called ends up calling itself (leading to a nested event subscribe before the correct event is fired). Added a check that the code called leading to the event is the same as the one we are waiting the event for.

2/ Added a BroadcastAndHold JSON-RPC call over HTTP. This is to bypass instability of the Websocket API.


silasdavis:
Thanks for the PR @ratranqu 

The CI is failing because there is a mock `transactor` that will need the `BroadcastTxAndHold` method added which I can easily do.

Splitting out `Hold` makes sense, and I don't see why not to have `BroadcastTxAndHold` as a method.

I think your solution is an improvement on what we have and is a pretty good heuristic, but I think we can still have the situation where an identical recursive call is made to the contract (with the same `CallData`) and we would still have the problem. I think the fundamental error is that we fire call events in post-order:

https://github.com/hyperledger/burrow/blob/master/manager/burrow-mint/evm/vm.go#L145

This allows it to catch any errors and output, so really it's a `AccCallComplete` event. We could introduce a `AccCallBegin` (or maybe `AccCall` and rename current `AccCall` to `AccReturn`...) that is triggered in pre-order. Then the first call event would be the one at the top of the stack, which is what `*AndHold` methods want... I think that might be a more general solution. Thoughts @benjaminbollen.

Which instability of the websocket API are you referring to? On 1337?

ratranqu:
@silasdavis: thx for adding the missing method in the tests. I only saw it after I did the PR, and didn't  find the time yet to update.

you're right on the identical call problem. I leave it to you guys to decide how best to handle.

re the instability of the websocket api, I only have circumstantial evidence for now. But I have noticed since 0.11.x (and up until 0.16.x at least) that when I use my libs (pure web sockets on 1337), after a while, the web sockets server on the burrow side stops broadcasting, and the only way for me to subscribe to events again is by killing the bc node I connect to and restarting it. Usually, this behaviour is triggered by killing a WS connection ungracefully (killing the connected socket due to time out for instance or killing the process). restarting the process alone without restarting the bc node did not help.

Additionally, when using the latest develop branch (cut from early July 2017), I noticed that burrow recovers from certain errors (can't remember which) when sending json-rpc calls through HTTP, but dies when the same requests are sent through WS. I suspect that some exception is not properly caught somewhere.

Anyhow, for now, quite busy on other stuff, so I don't hv a lot of time to go deeper in there for now.

silasdavis:
Closing in favour of: https://github.com/hyperledger/burrow/issues/715

PR #625, opened 2017-07-15
Protect WS endpoint against PANIC

ratranqu:
Currently, any Panic encountered during a WS call will kill the burrow process. This is not the case with http as gin-gonic protects against Panic (recovery.go).

This PR implements the same behaviour.


silasdavis:
Closing in favour of rebased: https://github.com/hyperledger/burrow/pull/717 thanks @ratranqu

PR #628, opened 2017-07-27
Fix incorrect link references in README.md

ghimire:


silasdavis:
I can see some small value in these changes, but more in using https-everywhere. If you would like to have them accepted please reopen from a branch into develop.

PR #637, opened 2017-08-25
Make sure we log somewhere if no config supplied

silasdavis:


core/config.go
```diff
@@ -107,6 +107,9 @@ func LoadServerConfig(chainId string, rootConfig *viper.Viper) (*server.ServerCo
 }
 
 func LoadLoggingConfigFromDo(do *definitions.Do) (*lconfig.LoggingConfig, error) {
+	if !do.Config.IsSet("logging") {
+		return nil, nil
+	}
 	loggingConfigMap := do.Config.GetStringMap("logging")
```
silasdavis: Without the check above we get an empty map that gives us a non-nil logging config which kills the logs. Before we were using `GetStringMap` this did not happen, but config structure was changed requiring this thus the lack of logs without an explicit config.

PR #644, opened 2017-08-31
Upgrade our CI/CD process

silasdavis:
Initially motivated by wanting to upgrade our images to use alpine 3.6 I have streamlined and improved the build process in the following ways:

- Upgraded our base image to alpine 3.6 and go 1.9.0
- Upgraded to CircleCI 2.0 (mostly to get access to docker >= 17.05, but this will be the version they support going forward)
- Changed our two-step build process + script to a single multistage build (this gives us the same small build image without juggling artefacts
- Dropped out dependence on Monax build and base images (defined in https://github.com/monax/monax). We now just depend on our own Dockerfile and a well maintained external images, so it should be easier to upgrade and get security fixes. 
- Enhanced our immutability guarantees around released images by:
  - Not pushing a release unless code version and tag version match
  - Pushing a separate `db-dev-${time}-${commit}` images from develop to avoid accidentally depending on a 'mutable pre-release'

Note that in order to get branch and tag filtering on CircleCI 2.0 I had to use their 'workflow' feature. We could potentially use this for other things, but frankly it was a bit of a pain to get it to do something as obvious as this setup is. Still it's nice to watch the workflow graph when it works...

I have been testing and iterating the circle process on my own fork. A commit to develop will trigger a build like this:
https://circleci.com/workflow-run/02e47d4f-dac1-4380-bce0-07d794e52f99

A tag is similar, but the build tool would tag it with the version in version.go (provided it the tag version matches).

Any other commit just tests:
https://circleci.com/workflow-run/6f4c8d9a-b828-472c-b74b-fd0917fab1bd

I have run https://github.com/monax/monax/blob/master/tests/test_jobs.sh against a chain running the image built from this and the tests pass.


compleatang:
LGTM 👍 

This will be much more sustainable and less "monax-y" going forward.

RobertDiebels:
Hey guys, I was hoping something like this would be pushed soon. Have you pushed the container image somewhere already?
I'd like to see if I can use it.

silasdavis:
Will be pushed shortly after this is merged and tagged on master: https://github.com/hyperledger/burrow/pull/646

RobertDiebels:
@silasdavis Great! Will follow that merge :D !

RobertDiebels:
@silasdavis Btw, I was reading the Dockerfile and this might be what you're looking for https://docs.docker.com/engine/reference/builder/#arg concerning the sharing of variables between build stages.

silasdavis:
@RobertDiebels Oh yes that's exactly what I was looking for, I'll put it into a patch release: https://github.com/hyperledger/burrow/pull/648 with some other tweaks. Thanks!

Incidentally now we are pushing distinct `db:${version}-dev-${date}-${commit}` builds you can create a chain with them like this:

```
monax chains make --version "quay.io/monax/db:$version-dev-$date-$commit" --chain-type simplechain yellow_chain
```

RobertDiebels:
@silasdavis no problem. Glad to help :D.

I'll take a look at the repo thanks! Any chance you'll also be pushing to Hyperledger's docker repo?

compleatang:
It would be good to push to HL's docker repo. We just need to track down who we talk to for access keys...

PR #662, opened 2017-12-29
Glide dev install patch 660

tommling:
**Summary:**
As per issue 660, fix the dependency missing issue when using glide-install. This is mainly a work around while burrow catches up to a slightly newer tendermint version.

**Why the pull request?**
Basically, v0.9.2 tries to pull
```
[WARN]	Unable to checkout github.com/tendermint/go-clist
[ERROR]	Error looking for github.com/tendermint/go-clist: Unable to get repository: Cloning into '/home/t/.glide/cache/src/https-github.com-tendermint-go-clist'...
ERROR: Repository not found.
```
but go-clist now lives in a subpackage of github.com/tendermint/tmlibz

Unfortunately, not all the packages were migrated so we have some issues with multiple versions i.e.:
```
[ERROR]	Unable to import from github.com/tendermint/go-p2p. Err: Import github.com/tendermint/tmlibs repeated with different versions '' and 'develop'
```

**What does it do?**
Simply points glide to package it needs.


silasdavis:
Thanks for this, but replacing with this: https://github.com/hyperledger/burrow/pull/672 for reasons detailed there

PR #665, opened 2018-01-08
Magic bytes to opcodes and comments

smblucker:
Convert magic byte values to constant opcode references and add comments in vm_test.go

[vm_test.go.zip](https://github.com/hyperledger/burrow/files/1612108/vm_test.go.zip)


silasdavis:
As we've discussed you'll re-open this against the feature branch

PR #666, opened 2018-01-09
Hypermarmot refactor

silasdavis:
# Refactor Summary:
For background see: https://lists.hyperledger.org/pipermail/hyperledger-tsc/2017-October/001163.html

About this PR:
**Note: rpc/v0 is currently out of action, I will resolve this with a future commit (possibly future PR)**

The feature/hypermarmot branch is even with develop/master apart from it containing this large commit: https://github.com/hyperledger/burrow/pull/666/commits/3ba7d190c2ece6b2baefb11b7bf0064b8ea5680c that pulls in the new vendor directory on which these commits have been rebased. This is to avoid third party code that we were acquiring through glide polluting this PR.

Key themes:
- Be a better EVM/execution library; more interfaces for key data structures and clearer boundaries
- Try to be a good Tendermint App rather than fight against it
- More immutability and type safety - ask if you want a mutable type, use stronger typing, offer more specific mutations (e.g. `MutableAccount.IncSequence()`)
- Pull monax tool functionality and tests into Burrow
- Large amounts of residual cleanup dating from original split with Tendermint
- Vendor dependencies for better reproducibility of builds

## Code hygiene
- Unified `Account` notion and introduced interface to support 
- Introduce composable State interfaces
- Use appropriate unsigned types for non-negative quantities
- Default to immutable types
- Move types to where their core functionality is defined (e.g. drop core/types and other delocalisation of types)
- Dial back punning of Tendermint types - but maintain some separation between execution, serialisation, and display types (for the same logical construct) where necessary (i.e. when they may reasonably change for different reasons)
 
- Communication between `TxCache`, `BlockCache`, and `State` strictly through 'State' interfaces rather than database sharing (prepares way for Merkle tree refactor)

## Project infrastructure 
- Transitioned from Glide to Dep
- Vendored dependencies
- Upgraded all dependencies

## Technical debt
- Removed manually mapped config
- Dropped pluggable modules notion
- Removed independent module versioning
- Dropped 'Pipe' notion and lazy loading
- Removed some unnecessary layers of indirection (interfaces)

## Accounts
Replaced the following structs with interfaces to encapsulate and allow immutability:
- Account
- Validator
- PrivateAccount
- Added `Concrete*` serialisation type
- Added 20 byte `Address` type (as opposed to `[]byte`) with various conveniences

## State
- account/state interfaces provide boundary over which state is accessed uniformly
- Introduced `Blockchain`, `Tip`, and `Root` to represent 'our' side of the blockchain state - that which we track, keep synchronised, and use in execution

## Execution
- `BatchExecutor` and `BatchCommitter` wrap `BlockCache` preparing for improved implementation
- `TxEncoder` and `TxDecoder` introduced to make possible to migrate to different binary encoding

## RPC
- Introduce central unified service.go
- Serialisation of `Result` via go-wire data mapper

### tm
- Rename to 'tm' in an effort to forget its origins and avoid confusion with our consensus engine/modules

## Tendermint
- Upgraded to 0.12.0 (and counting)
- Provided read-only view of node
- Implemented monax-keys interface backed PrivValidator (having requested change in Tendermint)
- Provide GenesisDoc in memory - derived from our GenesisDoc rather than passing around file
- Stopped exposing all possible config - will expose the bare minimum and coordinate those values

## Genesis
- `GenesisSpec` provides genesis templating and `monax chains make` type functionality, e.g.
```shell
# Chain one-liner to create keys according to spec, formulate GenesisDoc and config
burrow spec --participant-accounts 2 --full-accounts 1 | burrow configure --genesis-spec=- --validator-index 0 | burrow -c-
```
- Burrow -> TM genesis closed in via AppHash
- Genesis account types made compatible and convertible to/from `Account`/`PrivateAccount`

## Config
- Mapped automatically to/from `BurrowConfig` struct
- sources package providers ability to chain `ConfigProviders` (e.g. environment, file, XDG) together, to use a single file, split across multiple, or read directly from stdin.

## Events
- Serialisation via go-wire data mapper
- Introduce `AnyEventData` sum type
- Separate EVM and execution events
- Fixed empty cache bug

## Tx
- Serialise via data-mapper

## Serialisation
- Drop unnecessary tags and support TOML/JSON


account/bytecode.go
```diff
@@ -0,0 +1,36 @@
+package account
+
+import (
+	"encoding/json"
+
+	"github.com/tmthrgd/go-hex"
+)
+
+// TODO: write a simple lexer that prints the opcodes. Each byte is either an OpCode or part of the Word256 argument
+// to Push[1...32]
```
rotestein: Should create an issue for this when the pr is merged

account/keys.go
```diff
@@ -0,0 +1,61 @@
...
+
+type PublicKey struct {
+	crypto.PubKey `json:"unwrap"`
+}
+
+func PublicKeyFromPubKey(pubKey crypto.PubKey) PublicKey {
+	return PublicKey{PubKey: pubKey}
+}
+
+func PrivateKeyFromPrivKey(privKey crypto.PrivKey) PrivateKey {
```
rotestein: `PublicKeyFromPubKey` and `PrivateKeyFromPrivKey` are weird names. It's not obvious what they do without context. Suggest maybe `P*KeyFromCryptoP*Key` or maybe just omitting these functions since they add an extra layer of indirection without adding much.

account/account.go
```diff
@@ -41,64 +36,319 @@ func SignBytes(chainID string, o Signable) []byte {
...
+}
+
+func GetMutableAccount(getter Getter, address Address) (MutableAccount, error) {
+	acc, err := getter.GetAccount(address)
+	if err != nil {
+		return nil, err
+	}
+	// If we get get our own concreteAccountWrapper back we can save an unnecessary copy and just
+	// return since ConcreteAccount.Account() will have been used to produce it which will already
+	// have copied
```
rotestein: Is it a good idea to depend on this implementation detail of concreteAccountWrapper? Does it hurt anything to always pass acc to AsMutableAccount()?

keys/key_client.go
```diff
@@ -15,21 +15,48 @@
...
+
+// This mirrors "github.com/monax/keys/crypto/KeyType" but since we have no use for the struct here it seems simpler
+// to replicate rather than cop an import
+type KeyType string
+
+func (kt KeyType) String() string {
+	return string(kt)
 }
 
+const (
```
compleatang: ++ for adding secp2k1

test/server/scumbag_test.go
```diff
@@ -1,89 +0,0 @@
-// Copyright 2017 Monax Industries Limited
```
compleatang: RIP `scumbag` :(

genesis/deterministic_genesis.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func (dg *deterministicGenesis) GenesisDoc(numAccounts int, randBalance bool, minBalance uint64, numValidators int,
+	randBonded bool, minBonded int64) (*GenesisDoc, []acm.PrivateAccount) {
+
+	accounts := make([]Account, numAccounts)
+	privAccounts := make([]acm.PrivateAccount, numAccounts)
+	defaultPerms := permission.DefaultAccountPermissions
+	for i := 0; i < numAccounts; i++ {
+		account, privAccount := dg.Account(randBalance, minBalance)
```
compleatang: why `randBalance` ...? I'm not sure I follow the logic of that.

genesis/maker.go
```diff
@@ -61,20 +73,252 @@ func NewGenesisValidator(amount int64, name string, unbondToAddress []byte,
...
+				Address: addr,
+				Amount:  amt,
+			},
+		},
+	}
+	// [zr] why no index < 0 like in genDocAddAccount?
+}
+
+//-----------------------------------------------------------------------------
+// util functions
```
compleatang: We should review the below functions as I'm pretty sure they're redundant with other code or not needed.

genesis/spec/presets.go
```diff
@@ -0,0 +1,74 @@
...
+	"fmt"
+
+	"sort"
+
+	"github.com/hyperledger/burrow/permission"
+)
+
+// Files here can be used as starting points for building various 'chain types' but are otherwise
+// a fairly unprincipled collection of GenesisSpecs that we find useful in testing and development
+
```
compleatang: We should add root and developer account types here.

txs/tx_utils.go
```diff
@@ -37,65 +33,73 @@ func NewSendTx() *SendTx {
...
 	}
 
-	nonce := acc.Sequence + 1
+	nonce := acc.Sequence() + 1
 	return NewCallTxWithNonce(from, to, data, amt, gasLimit, fee, nonce), nil
 }
 
-func NewCallTxWithNonce(from crypto.PubKey, to, data []byte, amt, gasLimit, fee int64, nonce int) *CallTx {
-	addr := from.Address()
+func NewCallTxWithNonce(from acm.PublicKey, to *acm.Address, data []byte,
```
compleatang: `NewCallTxWithSequence` for consistency?

execution/block_cache.go
```diff
@@ -0,0 +1,386 @@
...
+	storage := iavl.NewIAVLTree(1024, db)
+	storage.Load(root)
+	return storage
+}
+
+var _ acm.StateWriter = &BlockCache{}
+
+var _ acm.StateIterable = &BlockCache{}
+
+// TODO: BlockCache badly needs a rewrite to remove database sharing with State and make it communicate using the
```
compleatang: Reasonable `help wanted` issue...?

permission/permissions.go
```diff
@@ -0,0 +1,190 @@
+// Copyright 2017 Monax Industries Limited
```
compleatang: New `perms` implementation much cleaner and more readable. ++

consensus/tendermint/config.go
```diff
@@ -1,177 +1,42 @@
...
 }
 
-//------------------------------------------------------------------------------
-// Tendermint defaults
-
-//
-// Contract
-//
-
-func (tmintConfig *TendermintConfig) AssertTendermintDefaults(chainId, workDir,
```
compleatang: I like this more opinionated take on running tm. It will significantly declutter (and de-confuse) our configs.

logging/config/sinks.go
```diff
@@ -187,29 +191,6 @@ func StderrOutput() *OutputConfig {
...
-
-func RemoteSyslogOutput(tag, remoteUrl string) *OutputConfig {
-	return &OutputConfig{
-		OutputType: Syslog,
-		SyslogConfig: &SyslogConfig{
-			Url: remoteUrl,
-			Tag: tag,
-		},
-	}
-}
```
aludvik: Does this commit remove support for logging to a file or syslog?

keys/key_client.go
```diff
@@ -15,21 +15,48 @@
...
+	KeyTypeEd25519Ripemd160         KeyType = "ed25519,ripemd160"
+	KeyTypeEd25519Ripemd160sha256           = "ed25519,ripemd160sha256"
+	KeyTypeEd25519Ripemd160sha3             = "ed25519,sha3"
+	KeyTypeSecp256k1Ripemd160               = "secp256k1,ripemd160"
+	KeyTypeSecp256k1Ripemd160sha256         = "secp256k1,ripemd160sha256"
+	KeyTypeSecp256k1Ripemd160sha3           = "secp256k1,sha3"
+	KeyTypeDefault                          = KeyTypeEd25519Ripemd160
```
rotestein: Is support for all of these types actually being added in this commit, or is this preliminary?

account/account.go
silasdavis: `AsMutableAccount` depends on same detail, but I agree better not to

account/keys.go
silasdavis: I agree on the names. I sort of intended the indirection though. I'd like a single place to hook any changes in dependency. Currently it's a simple wrapping but it could be a true mapping. I'm going to hold my nose and favour the full `PrivateKeyFromGoCryptoPrivKey`

account/account.go
```diff
@@ -41,64 +36,319 @@ func SignBytes(chainID string, o Signable) []byte {
...
+// The default immutable interface to an account
+type Account interface {
+	Addressable
+	// The value held by this account in terms of the chain-native token
+	Balance() uint64
+	// The EVM byte code held by this account (or equivalently, this contract)
+	Code() Bytecode
+	// The sequence number (or nonce) of this account, incremented each time a mutation of the
+	// Account is persisted to the blockchain state
+	Sequence() uint64
```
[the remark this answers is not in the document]
silasdavis: It's something of a cascade from having block heights as `uint64`, see my issue here: https://github.com/tendermint/tendermint/issues/731. `int` only takes us to 68 years! Clearly nonce can grow at faster linear rate to this, so makes sense to pull in.

account/account.go
```diff
@@ -41,64 +36,319 @@ func SignBytes(chainID string, o Signable) []byte {
...
+	Encode() []byte
+}
 
-	Permissions ptypes.AccountPermissions `json:"permissions"`
+type MutableAccount interface {
+	Account
+	// Set public key (needed for lazy initialisation), should also set the dependent address
+	SetPublicKey(pubKey PublicKey) MutableAccount
+	// Subtract amount from account balance (will panic if amount is greater than balance)
+	SubtractFromBalance(amount uint64) MutableAccount
```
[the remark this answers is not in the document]
silasdavis: The logic was that if we get here we have failed other checks and the desire to keep the interface simpler. However, it was the wrong reasoning and I agree with you.

account/private_account.go
```diff
@@ -0,0 +1,136 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package account
+
+import (
+	"fmt"
+
+	"github.com/tendermint/ed25519"
```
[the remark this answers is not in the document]
silasdavis: We should leave that to a separate PR

consensus/tendermint/config.go
silasdavis: Good, I think we will need to expose more configurability, but rather than just checking every possible changing value I'd like us to expose a higher level interface to tuning that works with Burrow

execution/block_cache.go
silasdavis: it's on roadmap yeah

genesis/deterministic_genesis.go
silasdavis: it's legacy from where this code was derived. It used to be truly random, I made is deterministically random. Really it doesn't need to be either, you could feasibly use it to do some automated testing over a range of seeds. I wouldn't have written it this way, and I possibly should have cleared it up when I rewrote parts but I wasn't clear on what tests were depending on so I gave option of picking some random values. I don't think it does much harm leaving it in for now.

genesis/maker.go
silasdavis: oops yeah that should be deleted, found its way back in during rebase.

genesis/spec/presets.go
silasdavis: my thinking was that the presets functionality was implemented in structure but still something of a stub, but yeah for parity/completeness I'll add those. Note that we can have any manner of compositional presets with multiple accounts or whatever.

txs/tx_utils.go
silasdavis: Casey you need to be stopped. Consistency this. Consistency that. It's globalisation gone mad. Go on then... There's a few other usages of nonce I'll change too.

keys/key_client.go
silasdavis: There is support for this in our 'reference' implementation of keys - I decided to import all these constants here in anticipation of using them at a future date.

keys/key_client.go
```diff
@@ -50,34 +123,59 @@ func NewBurrowKeyClient(rpcString string, logger logging_types.InfoTraceLogger)
 
 // Monax-keys client Sign requests the signature from BurrowKeysClient over rpc for the given
 // bytes to be signed and the address to sign them with.
-func (monaxKeys *monaxKeyClient) Sign(signBytesString string, signAddress []byte) (signature []byte, err error) {
+func (monaxKeys *monaxKeyClient) Sign(signAddress acm.Address, message []byte) (crypto.Signature, error) {
```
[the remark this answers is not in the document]
silasdavis: good point, I agree

logging/config/sinks.go
silasdavis: It is because we dropped our log15 dependency that was providing this. I would like to reintroduce syslog, but I can put file back in their now.

keys/key_client.go
```diff
@@ -15,21 +15,48 @@
 package keys
 
 import (
+	"bytes"
 	"encoding/hex"
 	"fmt"
 
+	acm "github.com/hyperledger/burrow/account"
```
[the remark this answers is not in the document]
silasdavis: I'm not a fan of it either particularly, though at least it's short. I used it because there was a pre-existing usage of it in a few places. Based on the fact previous authors use `dbm` for `go-db`, package `db` I sort of read it as 'database manager' and 'account manager'.

My two reasons for using an alias were:
1. To leave 'account' available for actual accounts - actually this doesn't happen a lot because there are usually `<something>Account`s.
2. Because I don't like `account.Account` as a type name, yet nothing came immediately to mind to replace it.

I don't think these are particularly strong reasons though I'd be tempted to pick another name for the `account` package if I did drop the alias. `state` perhaps, but a bit generic...

keys/key_client.go
```diff
@@ -39,6 +66,52 @@ type monaxKeyClient struct {
 	logger    logging_types.InfoTraceLogger
 }
 
+type monaxSigner struct {
```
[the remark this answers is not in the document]
silasdavis: I'll just drop, we're already in burrow

keys/key_client.go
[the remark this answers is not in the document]
silasdavis: FYI I'm going to do something about this, but in future single-purpose PR since it will touch almost every file

account/account.go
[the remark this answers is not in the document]
silasdavis: It's only adds overhead on systems that have 32-bit words, which IOT possibilities aside, is probably not that many. On 64-bit systems it's free.

Also I do not think it's unnecessary - suppose I hammer a chain using the same account as input with 100 txs per block - unusual but perfectly within the realm of load testing - then we run out of int32s in about 8 months. Worse than this is _if_ we have validators running on a mix of 32-bit and 64-bit architectures the 32-bit ones will die while the 64-bit ones won't. By forcing this as uint64 then we have parity across architectures - and the overhead is really minimal, just space not time really because we do minimal arithmetic on these. So I think this make sense right now, even if we don't _need_ it for height (even then - simulating a long-running chain - the disparity between archs is not nice)

account/account.go
```diff
@@ -284,22 +278,22 @@ func (caw concreteAccountWrapper) SetPublicKey(pubKey PublicKey) MutableAccount
 	return caw
 }
 
-func (caw concreteAccountWrapper) SubtractFromBalance(amount uint64) MutableAccount {
+func (caw concreteAccountWrapper) SubtractFromBalance(amount uint64) (MutableAccount, error) {
 	if amount > caw.Balance() {
-		panic(fmt.Errorf("insufficient funds: attempt to subtract %v from the balance of %s",
-			amount, caw.ConcreteAccount))
+		return nil, fmt.Errorf("insufficient funds: attempt to subtract %v from the balance of %s",
+			amount, caw.ConcreteAccount)
```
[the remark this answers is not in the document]
silasdavis: yeah, there were actually quite a few inherited panics that I've remove in this batch of changes - not quite sure why I kept them. This one I added myself - I guess a kind of 'bar' was set - so good you challenged on this.

account/crypto.go
```diff
@@ -0,0 +1,235 @@
...
+
+import (
+	"bytes"
+	"crypto/rand"
+	"crypto/sha256"
+	"fmt"
+	"io"
+
+	"github.com/tendermint/go-crypto"
+	"golang.org/x/crypto/ed25519"
```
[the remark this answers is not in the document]
silasdavis: No sure where we will end up with go-crypto... I think the aim will be to make it increasingly replaceable, but it may be that it does most of what we want. There's a few things it does I find a bit dubious, but we'll see...

execution/execution.go
```diff
@@ -781,7 +799,10 @@ func (exe *executor) Execute(tx txs.Tx) error {
 
 		// Good!
 		inAcc.IncSequence()
-		inAcc.SubtractFromBalance(value)
+		inAcc, err = inAcc.SubtractFromBalance(value)
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
silasdavis: I think they issue you may have stumbled upon, and I haven't found a good workaround yet, is that we are using Go.

genesis/maker.go
```diff
@@ -1,324 +0,0 @@
...
-	}
-
-	// convert pubkey hex strings to struct
-	pubKeys, err = pubKeyStringsToPubKeys(pubkeys)
-	if err != nil {
-		return
-	}
-
-	return pubKeys, amts, names, perms, setbits, nil
-}
```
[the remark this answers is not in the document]
silasdavis: 😢

silasdavis:
@rotestein if you have a chance to look at the changes in the last 4 commits then great, otherwise I'm looking to merge this today or early next week. If you're reading this an its merged I will review any comments and look to address them in a future PR to this feature branch. Thanks again for review.

genesis/deterministic_genesis.go
compleatang: Cool.

PR #671, opened 2018-01-18
Gofmt -s

robert-zaremba:
The repository contains some `gofmt -s`  issues. This changeset is an autofix of solve them and Makefile integration.

## Motivation and Context

For clarity in future pull-requests, the `gofmt -s` issues should be solved once and for all.

This change doesn't break the code, because it uses official `gofmt` to only do the code formatting.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,12 +28,12 @@ version:
 ### Formatting, linting and vetting
 
 # check the code for style standards; currently enforces go formatting.
-# display output first, then check for success	
+# display output first, then check for success
 .PHONY: check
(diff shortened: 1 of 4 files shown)


silasdavis:
This looks bot-like to me. If it is, please don't do this, if not we don't use rewrite rules so `-s` is no use here. Also this changes the meaning of `make check`

robert-zaremba:
Yes, it's mostly automatic change. 
`gofmt` is a great tool, why not taking full advantage of it?

PR #676, opened 2018-01-29
Reinstate RPC/V0

silasdavis:
This reinstates the V0 RPC; plugging it in to the new `Service` backend, with the intention of maintaining substantial wire-level compatibility (though since the case convention used for accounts has changed this may break the Javascript libraries downstream - the fix can be straight-forward but I will revisit when testing legacy-contracts).

This includes the following changes:
- Remove the restful URL interface to V0
- Drop `rpc.Result` and go-wire from the `Service` interface types and rpc/TM
- Improve the event subscriptions websocket code to make use of the JSON-RPC ID
- Drop the V0 mocked and 'scumbag' tests which were fragile and of limited usefulness
- Make use of the graceful `Shutdown(context.Context) error` supported by Go's `http.Server` since 1.8 in anticipation of replacing much of the hand-rolled server code (in particular when adding web3)
- Fixes tests for new vendored code


rpc/v0/codec.go
```diff
@@ -59,14 +59,14 @@ func (this *TCodec) Decode(v interface{}, r io.Reader) error {
...
-func (this *TCodec) DecodeBytes(v interface{}, bts []byte) error {
+func (codec *TCodec) DecodeBytes(v interface{}, bts []byte) error {
 	var err error
 	wire.ReadJSON(v, bts, &err)
 	return err
 }
 
 // Decode from a byte array pointer.
-func (this *TCodec) DecodeBytesPtr(v interface{}, bts []byte) error {
+func (codec *TCodec) DecodeBytesPtr(v interface{}, bts []byte) error {
```
compleatang: RIP `this` :)

txs/tx.go
```diff
@@ -114,11 +114,6 @@ type (
 		DecodeTx(txBytes []byte) (Tx, error)
 	}
 
-	// UnconfirmedTxs
```
compleatang: assuming since the endpoint is still available that this information can still be pulled? which has been helpful at times in debugging whether a mempool is queueing...

Dockerfile
```diff
@@ -33,4 +33,4 @@ EXPOSE 1337
 EXPOSE 46656
 EXPOSE 46657
 
-CMD [ "burrow", "serve" ]
+CMD [ "burrow" ]
```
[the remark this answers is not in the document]
silasdavis: The only thing is we are currently also providing a `burrow-client` binary here. Possibly we shouldn't. This docker file is also not much use on its own without being hydrated with a config. I'm tempted to just release Burrow as many cross-compiled native binaries using goreleaser (like with Hoard) and drop the dockerfile altogether. It's only really useful in the context of an orchestration system. I guess current version is runnable from environment variable. I think I'll leave things for now and return to this to make it a bit more functional.

rpc/service.go
```diff
@@ -81,50 +79,55 @@ type service struct {
 }
 
 var _ Service = &service{}
+var _ event.Subscribable = Service(nil)
 
-func NewService(state acm.StateIterable, eventEmitter event.Emitter, nameReg execution.NameRegIterable,
+func NewService(state acm.StateIterable, nameReg execution.NameRegIterable, eventEmitter event.Emitter,
```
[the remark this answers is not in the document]
silasdavis: It's more that it's being segregated than being elevated. Although it is currently stored in the same underlying state tree I didn't want the name reg interface to be part of the main state interface so it isn't part of the EVM `StateIterable`. I cam quite close to removing it altogether for the Hypermarmot refactor, but decided to do the work to keep it in, but I still see it as a candidate for removal - I don't think it makes sense as a core part of Burrow.

txs/tx.go
silasdavis: Oh yes it is this, but this in an RPC view model and just didn't belong here - it was unused since refactor but noticed it during these changes.

rpc/service.go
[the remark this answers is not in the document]
compleatang: It's not used enough to be a core feature. it's also very easily implemented in smart contracts. The only advantage is that we can ping DOUG in the nameReg and never need to store a single address to get connect into a system of smart contracts.

txs/tx.go
compleatang: 👍

PR #681, opened 2018-02-07
Add VM test to check return of msg.sender

silasdavis:
Courtesy of @guoger (thanks) - with some updates to use initialised contract code

### DIFF
--- a/execution/evm/vm_test.go
+++ b/execution/evm/vm_test.go
@@ -31,14 +31,16 @@ import (
 	. "github.com/hyperledger/burrow/execution/evm/asm"
 	. "github.com/hyperledger/burrow/execution/evm/asm/bc"
 	evm_events "github.com/hyperledger/burrow/execution/evm/events"
-	"github.com/hyperledger/burrow/logging/lifecycle"
 	"github.com/hyperledger/burrow/logging/loggers"
 	"github.com/hyperledger/burrow/permission"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/crypto/ripemd160"
 )
 
-var logger, _ = lifecycle.NewStdErrLogger()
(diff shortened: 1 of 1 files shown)


execution/evm/vm_test.go
```diff
@@ -286,6 +287,49 @@ func TestMemoryBounds(t *testing.T) {
...
+
+			contract SimpleStorage {
+		                function get() public constant returns (address) {
+		        	        return msg.sender;
+		    	        }
+			}
+	*/
+
+	// This bytecode is compiled from Solidity contract above using remix.ethereum.org online compiler
+	code, err := hex.DecodeString("6060604052341561000f57600080fd5b60ca8061001d6000396000f300606060405260043610603f576000357c0100000000000000000000000000000000000000000000000000000000900463ffffffff1680636d4ce63c146044575b600080fd5b3415604e57600080fd5b60546096565b604051808273ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff16815260200191505060405180910390f35b6000339050905600a165627a7a72305820b9ebf49535372094ae88f56d9ad18f2a79c146c8f56e7ef33b9402924045071e0029")
```
guoger: for pure testing purpose, we probably could directly use **runtime bytecode** instead of **deploy bytecode** here, so that we don't necessarily need to run evm twice

silasdavis: There's something to be said for using the runtime code, but it does put it one more step removed from the compiled solidity (i.e. to verify external to Burrow you'll need to run a compiler and deploy a contract to get the same bytecode - not a huge deal), but more than that I think I like the fact it documents this aspect of how Ethereum contract deployment works - which is not very widely explained - the fact you came up against this is perhaps evidence of this.

I'll split the lines. I also would like to avoid hex bytecode and use the constant symbols in these tests. For this I need to write a simple lexer. If we have human-readable bytecode we can try and have more tests that are divorced from solidity without being overly simple.

PR #688, opened 2018-02-18
Fix accidental compositing of logging config

silasdavis:
We use the default config as a base (so you only need to set non-default values) but in the case of the recursive logging structure this means that when you explicitly set a logging config it gets merged into the default that just outputs unfiltered thus breaking custom logging config. This fixes that.


compleatang:
LGTM

PR #695, opened 2018-02-26
Run integration tests from Bosmarmot

silasdavis:
Runs integration tests from Bosmarmot pinned to a particular version

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,10 @@ SHELL := /bin/bash
 REPO := $(shell pwd)
 GOFILES_NOVENDOR := $(shell find ${REPO} -type f -name '*.go' -not -path "${REPO}/vendor/*")
 PACKAGES_NOVENDOR := $(shell go list ./... | grep -vF /vendor/)
-COMMIT_SHA := $(shell echo `git rev-parse --short --verify HEAD`)
+# Bosmarmot integration testing
+BOSMARMOT_PROJECT := github.com/monax/bosmarmot
+BOSMARMOT_GOPATH := ${REPO}/.gopath_bos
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -91,27 +94,36 @@ build:	check build_db build_client
...
 .PHONY: build_db
 build_db:
-	go build -o ${REPO}/target/burrow-${COMMIT_SHA} ./cmd/burrow
+	go build -o ${REPO}/bin/burrow ./cmd/burrow
 
 # build burrow-client
 .PHONY: build_client
 build_client:
-	go build -o ${REPO}/target/burrow-client-${COMMIT_SHA} ./client/cmd/burrow-client
+	go build -o ${REPO}/bin/burrow-client ./client/cmd/burrow-client
```
zramsay: ++ for using bin

zramsay: would be good to have an `make install` target though

silasdavis: I did wonder about that, but I sort of feel like either we do OS distribution properly with packages appropriate for each OS/distribution or we don't bother at all. If `make install` is just copying `burrow*` to `/usr/local/bin` then it doesn't provide much value and on the contrary it risks obfuscating the fact that we are dealing with a statically compiled Go binary that you can just run from any location with the same arch.

I think down the line (probably before next release) I'd like to support https://goreleaser.com/. Initially just for cross-compiling, but it does have integration to proper packaging. It's just I don't really want to get (back) into supporting packaging - it's not worth our time.

zramsay: oh, i missed (originally) the removal of the `-${COMMIT_SHA}` whichhelps

PR #698, opened 2018-02-26
WIP: Implement Revert Opcode

smblucker:
These changes represent a first attempt at implementing the REVERT opcode (0xFD) into burrow's EVM; this includes assigning its hex and string value, writing its case definition, writing checks in relevant case definitions, and writing a test.

### DIFF
--- a/execution/evm/vm.go
+++ b/execution/evm/vm.go
@@ -45,6 +45,7 @@ var (
 	ErrDataStackUnderflow     = errors.New("Data stack underflow")
 	ErrInvalidContract        = errors.New("Invalid contract")
 	ErrNativeContractCodeCopy = errors.New("Tried to copy native contract code")
+	ErrExecutionReverted      = errors.New("Execution reverted")
 )
 
 type ErrPermission struct {
@@ -759,6 +760,10 @@ func (vm *VM) call(caller, callee acm.MutableAccount, code, input []byte, value 
 				stack.Push(newAccount.Address().Word256())
 			}
 
+			if err_ == ErrExecutionReverted {
+				return ret, nil
+			}
+
 		case CALL, CALLCODE, DELEGATECALL: // 0xF1, 0xF2, 0xF4
 			if !HasPermission(vm.state, callee, permission.Call) {
 				return nil, ErrPermission{"call"}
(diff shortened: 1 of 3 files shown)


execution/evm/vm.go
```diff
@@ -921,6 +926,10 @@ func (vm *VM) call(caller, callee acm.MutableAccount, code, input []byte, value
 				}
 			}
 
+			if err == ErrExecutionReverted {
```
silasdavis: This should be within the block above containing line 914, where the error is checked for readability. Note on develop I have renamed this `err` (which shadows the return value `err`!) to `callErr`, when you rebase against develop (which you need to do) you'll need to change that.

execution/evm/vm_test.go
```diff
@@ -158,6 +158,26 @@ func TestSubcurrency(t *testing.T) {
...
+func TestRevert(t *testing.T) {
+	ourVm := NewVM(newAppState(), DefaultDynamicMemoryProvider, newParams(), acm.ZeroAddress, nil, logger)
+
+	// Create accounts
+	account1 := newAccount(1)
+	account2 := newAccount(1, 0, 1)
+
+	var gas uint64 = 100000
+
+	bytecode := MustSplice(PUSH32, 0x72, 0x65, 0x76, 0x65, 0x72, 0x74, 0x20, 0x6D, 0x65, 0x73, 0x73, 0x61,
```
silasdavis: Cool...  It this bytecode the testcase from the EIP? If so can you add a comment with a link and reference to EIP-140

execution/evm/vm.go
```diff
@@ -936,6 +945,19 @@ func (vm *VM) call(caller, callee acm.MutableAccount, code, input []byte, value
...
 
+		case REVERT: // 0xFD
+			offset, size := stack.Pop64(), stack.Pop64()
+			output, memErr := memory.Read(offset, size)
+			if memErr != nil {
+				vm.Debugf(" => Memory err: %s", memErr)
+				return nil, firstErr(err, ErrMemoryOutOfBounds)
+			}
+
+			outputMessage := output[offset:size]
```
silasdavis: This is unnecessary, remove and just return `output` which is the correct slice already (as in `RETURN`)

execution/evm/vm.go
smblucker: You renamed err (in entire case definition for CALL, DELEGATECALL, and CALLCODE) to callErr, or you changed ErrExecutionReverted to callErr? Not sure what to do here other than move the above check to block beginning at LN# 914

smblucker: I will assume you meant you renamed err and not ErrExecutionReverted and push changes

silasdavis:
Merging into feature branch new-opcodes

PR #699, opened 2018-02-26
 Support creating contracts with Transact by accepting empty address

silasdavis:
This restores correct behaviour for JS libs, making this smoke test pass: https://github.com/monax/bosmarmot/blob/master/test/test.js.

This PR is stacked on top of: https://github.com/hyperledger/burrow/pull/695 so merge that first.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,8 +8,8 @@
 
 SHELL := /bin/bash
 REPO := $(shell pwd)
-GOFILES_NOVENDOR := $(shell find ${REPO} -type f -name '*.go' -not -path "${REPO}/vendor/*")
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -91,27 +94,36 @@ build:	check build_db build_client
 
 # build all targets in github.com/hyperledger/burrow with checks for race conditions
 .PHONY: build_race
-build_race:	check build_race_db build_race_client build_race_keys
+build_race:	check build_race_db build_race_clien
```
dennismckinnon: build_race_clien(t?) did you delete one too many letters?

PR #707, opened 2018-03-04
Support of embedded keys pair for tendermint private validator inside burrow instead of monax.keys

### LINKED ISSUE #702: [Keys] Support in-burrow crypto API instead of external monax.keys REST API
Would be nice to use burrow without any external crypto API REST server call like monax.keys.

There is already everything in MockKeyClient and tendermint/go-crypto to do it: allow an entry point in config API to override default monax.keys dependency within kernel creation:

1. Make ValidatorAddress not mandatory
2. Create tendermint.NewNode with a new burrow privValidator API
3. New entry in BurrowConfig that wrap Keys or new InMemoryCrypto struct
3. Support flexible API to acccess node private key in InMemoryCrypto

Example to be discussed:
```go
type BurrowConfig struct {
	// ValidatorAddress *acm.Address =>  Moved to Keys
	GenesisDoc       *genesis.GenesisDoc
	Tendermint       *tendermint.BurrowTendermintConfig
        // Keys             *keys.KeysConfig ==> Moved to CryptoConfig and Keys Server
	Crypto           *CryptoConfig // New top crypto config
	RPC              *rpc.RPCConfig
	Logging          *logging_config.LoggingConfig
}

type CryptoConfig struct {
	KeysServer  *keys.KeysConfig // If the user wants to keep current behavior
	InMemoryCrypto *InMemoryCrypto // If user wants in-node crypto
}

type InMemoryCrypto struct {
	Address    acm.Address
	PrivateKey [ed25519.PrivateKeySize]byte
	PublicKey  []byte
} 

type KeysConfig struct {
	URL string,
        ValidatorAddress *acm.Address
}

```
I think it is usefull for user who want to use only burrow and tendermint while minimizing REST invocations.

Thank you for your review, if you are ok with this idea I may submit a PR.

_like burrow, keep it up!_

phymbert:
This PR allows you to configure inmemory crypto validator if you do not want to depend on an external keys API to sign messages for private tendermint validator:

By API:
```go
        conf := config.DefaultBurrowConfig()
...
	conf.Crypto.KeysServer = nil // Disable external call to sign and load public key
        // Generate keys OR load your own pair programatically
	conf.Crypto.InMemoryCrypto, err = inmemory.DefaultInMemoryCryptoConfig() 
...
```

Or by Config (not **safest** approach to store private key in file) - example:
```toml
[Crypto]
  [Crypto.InMemoryCrypto]
     PrivateKey = "{\"type\":\"ed25519\",\"data\":\"YOUR PRIVATE KEY\"}"
     PublicKey = "{\"type\":\"ed25519\",\"data\":\"2e4bbbffad9325db4e18725c6f31234ee303ac43f8ff4bee167b6e83a4c49197\"}"
```

Fixes #702


silasdavis:
Thanks for this. I'll reply now rather than waiting until I've fully considered everything:

1. I'm inclined to include this functionality - but not expose it in config (yet) - we need a way of splitting up config in general and I think this probably encourages behaviour we don't want to encourage (i.e. keys in config file). I think local keys need to be in a properly permissioned separate file and for there to at least a warning if they are not.

2. I'm inclined to butcher some of what you have done and rework it so that we have something like a self-hosted set of keys. We will shortly be adding 'co-located signing' as in you can get Burrow to sign transactions if it has access to the signing key via keys. It might make most sense to pull in monax-keys wholesale for this - or at least a library portion of it to just make it 'something Burrow can do'.

3. We need to support `PrivateKey` not just ed25519 keys

phymbert:
Thank you for your reply. I agree with all 3 points above, feel free to rework this small contribution  if needed.
I am pleased to help if required, for example I may recode private key access to an external file.
We are waiting this feature for my project, happy to see embedded mona-keys inside burrow the soonest.

silasdavis:
@phymbert just to let you know we are going to add burrow-hosted keys as a feature - this can support this use case and also with address-based delegated signing that I am working on also allows for running Burrow as a signing delegate - useful as a local 'wallet' or in the cloud behind a service.

b00f:
Thanks @phymbert for embedding monax.key to burrow. I'd like to suggest if it is possible using an interface pattern for signer object. Burrow can easily implement any curve in a modular way, not only ed25519.  Look here: https://github.com/hyperledger/burrow/pull/707/files#diff-06dc382f674b6310ff1f5382daad00b7

silasdavis:
We now have support for your use case and more with: https://github.com/hyperledger/burrow/pull/776

Issue #713, opened 2018-03-07
High cpu load and performance problem


pascalverlinden:
**Version info**
Monax CLI Version: 0.17.0 & 0.18.0
Monax db 0.16.3
docker  17.03.0-ce 
OS: Ubuntu 14.04.5 LTS

**Description**
Creating many contracts leads to high CPU load and performance problems with the chain. After stopping the chain the CPU usage drops, but then it is no longer possible to start the chain as this leads to an error "panic: Paniced on a Sanity Check: Expected storeHeight (1867) and stateHeight (1866) to match", which seems to be a problem with Tendermint ( see https://github.com/tendermint/tendermint/issues/388 ). There has been a fix in Tendermint (version 0.9) but this version is not used yet by monax, apparently.

There might be a relation with the following issue: https://github.com/hyperledger/burrow/issues/611

**Steps to reproduce**
Example code is available in https://github.com/pascalverlinden/perfo

joeriexelmans:
Updated example code to work with latest version of burrow and bosmarmot tooling:
https://github.com/pascalverlinden/perfo/tree/latest-version

Performance issues are still present:
- After creating 80 contracts of type 'D', and then doing nothing, CPU usage of burrow process remains at 20%. After there are 160 contracts on the chain, CPU usage is 40%.
- Latency of creating a new contract goes from 1s to 2s after about 60 contracts.
- CPU usage is close to 100% while sending a single transaction per block.

silasdavis:
## Profiling
Here is a 30 second CPU profile from a fresh train humming along at < 1% CPU on my machine:

```
(pprof) top 40
Showing nodes accounting for 730ms, 96.05% of 760ms total
Showing top 40 nodes out of 233
      flat  flat%   sum%        cum   cum%
      70ms  9.21%  9.21%       70ms  9.21%  runtime.futex
      70ms  9.21% 18.42%       70ms  9.21%  runtime.usleep
      60ms  7.89% 26.32%       80ms 10.53%  github.com/hyperledger/burrow/vendor/github.com/tendermint/ed25519/edwards25519.FeMul
      60ms  7.89% 34.21%       60ms  7.89%  github.com/hyperledger/burrow/vendor/golang.org/x/crypto/ripemd160._Block
      50ms  6.58% 40.79%       50ms  6.58%  github.com/hyperledger/burrow/vendor/github.com/tendermint/ed25519/edwards25519.feSquare
[...]

PR #718, opened 2018-03-14
Hotfix BlockCache accumulating writes when it shouldn't

silasdavis:
This is a near-term fix for #713. Rather than accumulating a cache across block we just maintain it while a single block of transactions are being applied. This does not logically effect the result of a block.


silasdavis:
This is a near-term fix for https://github.com/hyperledger/burrow/issues/713

PR #720, opened 2018-03-15
Fix the Table of Contents in README.md

qpakzk:
First, synchronize headers with contents' name.
Equalize uppercase or lowercase letters with headers. Add a question mark.

Second, fix the link of "What is Burrow ?", which didn't work.

Third, change the loction of  "Future work" and "License".
Because the text has a different order with the table of contents.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,14 +9,14 @@ Hyperledger Burrow is a permissioned Ethereum smart-contract blockchain node bui
 
 ## Table of Contents
 
-- [What is burrow](#what-is-burrow)
+- [What is Burrow ?](#what-is-burrow-)
 - [Installation](#installation)
-- [For developers](#for-developers)
+- [For Developers](#for-developers)
 - [Usage](#usage)
 - [Configuration](#configuration)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -9,14 +9,14 @@ Hyperledger Burrow is a permissioned Ethereum smart-contract blockchain node bui
 
 ## Table of Contents
 
-- [What is burrow](#what-is-burrow)
+- [What is Burrow ?](#what-is-burrow-)
 - [Installation](#installation)
-- [For developers](#for-developers)
+- [For Developers](#for-developers)
```
silasdavis: We are using drop case on other words elsewhere, omit this and I will accept

zramsay:
most if not all of this was fixed on develop in #700

README.md
qpakzk: Do you want to undo, otherwise delete "[For developers](#for-developers)"?

zramsay: @qpakzk see my comment below, and PR #700

PR #721, opened 2018-03-15
State improvements

silasdavis:
This PR provides a number of state-related fixes and improvements:

- Reintroduces the saving and loading on startup of Burrow state to disk via go-db with blockchain and execution state stored separately but synchronised by the state root hash
- Remove poorly implemented `BlockCache` and `TxCache` in favour of a unified `StateCache` that does not access database
- Provides a separate `NameRegCache`
- Localises database access to `execution/state.go`
- Implement state writes/reads through the `account/state.go` interfaces only
- Upgrades our merkle tree to the one being maintained in https://github.com/tendermint/iavl (though not the latest version because it depends on a newer tmlibs version than the one required by our version of Tendermint)
- Provides versioned snapshots of state for every block - laying groundwork for doing something similar for events. Currently we are not exposing this versioned access
- Fixes some issues with test timings by subscribing to the account input event instead of new block

**Currently outstanding**
- Unit tests for `StateCache`, `NameRegCache`, and `State`
- StorageRoot is no longer being set on `ConcreteAccount`. We are using a single tree and a safer mechanism for database access, but we are not storing a tree per account so don't have easy access to a storage root for _each_ account. We can either scrap this, store a multitude of trees, or compute it on demand
- Upgrading to the newest IAVL tree - possibly we should depend on our own fork where we can vendor (or remove) the problematic dependencies as well as add some useful features (like the ability to store a forest of trees in the same database and load the roots).
- Deal with intermittent failure affecting `TestWSCall*` tests

Hoping that @smblucker @seanyoung can take a look at these issues


account/state.go
```diff
@@ -63,3 +63,11 @@ type StateWriter interface {
 	Updater
 	StorageSetter
 }
+
+type IterableStateWriter interface {
```
silasdavis: Relates to Cache which has `IterateAccounts` etc

execution/execution.go
```diff
@@ -32,6 +32,9 @@ import (
 	"github.com/hyperledger/burrow/txs"
 )
 
+// TODO
+const GasLimit = uint64(1000000)
```
compleatang: we should be exposing this as a genesis config option. not now, but for later.

execution/evm/vm.go
```diff
@@ -72,7 +72,7 @@ type VM struct {
 	memoryProvider func() Memory
 	params         Params
 	origin         acm.Address
-	txid           []byte
+	txHash         []byte
```
compleatang: better name...

execution/execution.go
silasdavis: yes, I was planning to address this and other things with the ValidatorTx changes

silasdavis:
Merging now, we can add some more scrutiny in the merge from develop

PR #726, opened 2018-04-02
T1 network stabilisation

silasdavis:
This is stacked on: https://github.com/hyperledger/burrow/pull/721 so merge that first.


silasdavis:
DCO thing seems to be broken from earlier rebase

PR #727, opened 2018-04-03
State refresh logging and debug

silasdavis:
This PR aims to stabilise Burrow on our large real-world set of contracts exercising a large portion of EVM and SNative functionality. In particular it:

- Implements EIP 150 so that EVM produced by recent (tested with solc 0.4.21) passes expected gas to child call frames
- Adds overflow safety checking when producing 64-bit integers from 256-bit EVM words
- Ensures that we refresh our mutable pointers to immutable account state after calls and creation (else we may not get latest snative permissions or other account state altered down the call stack)
- Massively optimises logging particularly providing a fast past for logging that does not dominate CPU usage
- Also refactors our `Logger` to be a somewhat simpler struct wrapper around two go-kit loggers
- Adds debug profiling server options and makes EVM-level debug accessible via an options system
- Various Go usage improves (e.g. dropping singleton interfaces)
- Adds a few more tests to `state.Cache` (more ought to be added)
- Adds git commit version to `burrow --version` when built with `make build_db` via `-ldflags "-X ..."`


client/node_client.go
```diff
@@ -41,7 +40,7 @@ type NodeClient interface {
 	ListValidators() (blockHeight uint64, bondedValidators, unbondingValidators []acm.Validator, err error)
 
 	// Logging context for this NodeClient
-	Logger() logging_types.InfoTraceLogger
+	Logger() *logging.Logger
```
compleatang: Lot cleaner.

execution/evm/vm.go
```diff
@@ -1128,3 +1124,28 @@ func transfer(from, to acm.MutableAccount, amount uint64) error {
 	}
 	return nil
 }
+
+// Dump the bytecode being sent to the EVM in the current working directory
+func dumpTokens(txHash []byte, caller, callee acm.Account, code []byte) {
```
compleatang: What's this useful for?

execution/evm/native.go
```diff
@@ -73,13 +74,14 @@ func ecrecoverFunc(state State, caller *acm.Account, input []byte, gas *int64) (
 	recovered, err := secp256k1.RecoverPubkey(hash, sig)
 	if err != nil {
 		return nil, err
+OH NO STOCASTIC CAT CODING!!!!
```
compleatang: 🐱

execution/evm/vm.go
silasdavis: Provides human-readable bytecode for each tx that is committed - can be convenient for comparing with other generated bytecode and can be clearer than the line-by-line opcode debugging - which also only prints the path that was executed rather than everything. That also doesn't have the tx and callee so obviously. Ultimately this could be combined/enhanced into a single debug, but I found both of these useful so rather than killing the debug I've made them available via options that seems like a fairly clean way of keeping them in the codebase.

execution/evm/native.go
silasdavis: Lol you know what that was actually an accidental copypasta from a discussion I was having with @dennismckinnon ... occasionally happens via trackpad... So an accidental code addition referring to a class of accidental code additions.

b00f:
Singleton  is not a good design pattern and it has some bottlenecks. But I think in burrow we can use this pattern for logger system. In this case having **thread-safe singleton logger** will reduce the code and make logger to be accessible everywhere. Also it's not necessary to pass the logger instance between packages.

PR #728, opened 2018-04-03
Mempool sequence numbers and address based transact

### LINKED ISSUE #715: Ensure TransactAndHold waits for the root AccountCall event
This issue is extracted from: https://github.com/hyperledger/burrow/pull/624.

`TransactAndHold` subscribes to the AccountCall event here: 

https://github.com/hyperledger/burrow/blob/develop/execution/transactor.go#L243

This waits for an AccountCall event that has a `TxHash` matching the broadcast transaction. However any nested call will also post an AccountCall event and since they are fired in post-order nested calls will be the earlier event. We need to wait for the root (externally generated) AccountCall event to be received. To do this we can attach the call depth (already tracked in the VM) to the event (and event tags) and when subscribing ensure we are at call depth 0 (the root frame).

The PR to solve this should include a test in execution_test.go that demonstrates that subscription to AccountCall can yield the wrong call event in the current implementation. We do not necessarily need to replicate at the integration test level against `TransactAndHold`

silasdavis:
This PR introduces address-based signing for the Transactor initially exposed via the V0 RPC. It includes a number of refactorings along the way relating to transaction signing, mempool state, sequence numbers, and transaction formation, including:

- Use of `AddressableSigner` in most places `PrivateAccount` was used abstracting over key client based signing and private key signing
- Refactoring transactions to capture all signing under a single method that is added to the `Tx` interface (breaking out txs into their own files for clarity)
- Adds some additional state tests
- Factors out integration test logic and uses this to add sorely needed v0 integration tests similar to tm integration test and identically wrapped
- Adds simple benchmark for transact
- Fixes #715 ensuring transact handles recursive calls correctly
- Enables access to the mempool as a `state.AccountGetter` so that `Transactor` can optimistically assign sequence numbers as a delegated signer
- Enables address based signing via the `InputAddress` field on various of the v0 methods while maintaining backward-compatibility by keeping `PrivKey`. Both options share the same refactored `Transactor` by using the `InputAccount` abstraction.
- Ensures that a mempool backlog produced by the transactor spanning many blocks will succeed if the transactions are compatible by committing state before reseting check cache


blockchain/blockchain.go
```diff
@@ -31,15 +31,15 @@ var stateKey = []byte("BlockchainState")
...
-	ChainID() string
 	// GenesisHash precomputed from GenesisDoc
 	GenesisHash() []byte
 	GenesisDoc() genesis.GenesisDoc
 }
 
 // Immutable pointer to the current tip of the blockchain
 type Tip interface {
+	// ChainID precomputed from GenesisDoc
+	ChainID() string
```
compleatang: Curious why we're moving `ChainID` from root to tip; its static so not really sure why we are passing it around with each block. Perhaps better safety would be my assumption....?

execution/transactor.go
```diff
@@ -224,17 +185,17 @@ func (trans *transactor) Transact(privKey []byte, address *acm.Address, data []b
 	}
 
 	// Got ourselves a tx.
-	txS, errS := trans.SignTx(tx, []acm.PrivateAccount{pa})
-	if errS != nil {
-		return nil, errS
+	err = tx.Sign(trans.tip.ChainID(), inputAccount)
```
compleatang: Ah, so we integrate ChainID here in the signer, that's why you've moved it to tip, no?

txs/names.go
```diff
@@ -31,22 +27,8 @@ var (
 
 	MaxNameLength = 64
 	MaxDataLength = 1 << 16
```
compleatang: Curious what this means?

txs/rebond_tx.go
```diff
@@ -0,0 +1,55 @@
+package txs
```
compleatang: Are these bond_tx and rebond_tx actually reimplementing validator changes in preparation for validator_tx or are they just getting back and mocked?

blockchain/blockchain.go
silasdavis: Though you grab tip immutably with `Tip()` so there is a copy when you pull a tip explicitly, but mostly we are just accessing the single instance embedded in `blockchain` - both when accessing values or doing a `Commit()`. Having said that I do agree with you that it seems more logical in `Root` as a static value. The reason I moved it is with the addition of `ChainID` `Transactor` need only ask for a `Tip` rather than a whole `Blockchain`. I could have just added `ChainID` as an argument to `Transactor`. Let me think about that, maybe that is a nicer way to go... A counter-argument is if you want `Tip` to be the chain information you need to sign at a particular height, then it seems like `ChainID` belongs with other tip data.

txs/names.go
compleatang: i thought i remembered it was 2^16. was just weirded out by the go syntax. that's a lot of bytes....

blockchain/blockchain.go
compleatang: Thanks. Realized 75% thru the review why.... Should have removed the comment. But always good to learn!

txs/rebond_tx.go
compleatang: cool

PR #731, opened 2018-04-04
fix key management on known accounts; add chain name flags

compleatang:
this PR integrates two changes:

* the addition of a `--chain-name` flag to both `burrow configure` and `burrow spec`. since we write the output via both commands, it should be overwrite-able by both commands. if we want to only allow the flag to exist in one place it should go in `burrow configure` rather than in burrow spec.
* fixing of a problem when the genesisSpec was given known accounts. the result of those known account|pubKeys were not properly passed through the parser.

### DIFF
--- a/genesis/spec/genesis_spec.go
+++ b/genesis/spec/genesis_spec.go
@@ -127,11 +127,12 @@ func (ta TemplateAccount) RealisePubKeyAndAddress(keyClient keys.KeyClient) (pub
 			return
 		}
 	} else {
-		address := ta.PublicKey.Address()
+		address = ta.PublicKey.Address()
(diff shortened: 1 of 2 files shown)


genesis/spec/genesis_spec.go
```diff
@@ -127,11 +127,12 @@ func (ta TemplateAccount) RealisePubKeyAndAddress(keyClient keys.KeyClient) (pub
 			return
 		}
 	} else {
-		address := ta.PublicKey.Address()
```
silasdavis: oh shadowing......

silasdavis:
Thanks!

PR #732, opened 2018-04-06
Add install_db target and dirty hash

silasdavis:
Add install_db make target and a hash summarising any tracked changes in
the working directory that will be added to burrow --version output when
burrow is build with a dirty working dir - useful for distinguishing
builds against uncommitted code.


silasdavis:
`git describe --tags` courtesy of @seanyoung

PR #737, opened 2018-04-10
Upgrade to tendermint 18.0.0

b00f:
Changes:

* Upgrade to tendermint 18
* remove validators from blockchain object (it's needed to show validators per round)
* Rename Validator to GenValidator in genesis.go for better code understanding
* Remove vendor folder from project (it's not necessary)
* removing json tags


compleatang:
Hi @b00f thanks for this! Currently we want to ensure that burrow remains go gettable so until dep is baked into go get we will keep the vendor directory. 

We're happy to review your other contributions when your ready and the tests are green.

silasdavis:
Thanks for taking an interest in Burrow. I was planning to upgrade Tendermint (again) soon, but had been stabilising develop for a network deployment first.

I'd love to review your changes related specifically to the Tendermint upgrade, but I'd need you to submit them in isolation.

To respond to your other state changes:

> remove validators from blockchain object (it's needed to show validators per round)

I do not understand your purpose here, but in any case the validator set on the blockchain struct is currently a stub for re-introducing some form of validator set changes (aka validator bonding). It will be mutated by transactions with an appropriate permission. This is a change coming soon.

> Rename Validator to GenValidator in genesis.go for better code understanding

Why does this give better understanding. I'd take some convincing that we need a cosmetic change like this things other tooling and deployments depend on this name. I had also opted to remove tags for the JSON name as unnecessary busywork so this would also need justification.

> Remove vendor folder from project (it's not necessary)

Dep has golang on the way to being a grown up language with respect to dependencies but it's not perfect. We relatively recently decided to vendor our dependencies and I'm not inclined to change it again. Having vendor included serves two purposes: 1) we can be go-gettable, and 2) where we are used as a library we can maintain our own independent set of dependencies. Our licensing is also more clearly auditable with vendor checked in.

It would be good to chat if you'd like to contribute here: https://chat.hyperledger.org/channel/burrow-contributors. I have a draft roadmap that I would like to share, but have been heads-down trying to stabilise Burrow after a major refactor.

b00f:
Upgrading to tendermint 18 is not a big deal as I know tendermint 16 and 17 has issues (17 had memory leak). I think we can have this PR open until you decide to upgrade tendermint.

I think validator-set should not be part of blockchain structure. Maybe Tip is better place to track validators after bondin or unbondig. In each round validator-set can be changed, so we need to get validator set for each block in every height separately. Currently burrow just shows validators in the latest height.

For me as a person who check the code for the first time, I got confused between genesis validator structure and acm.validator. This is the reason I renamed it for better code understanding. Other structure like Account can have Gen prefix. But if you think it'snot a good idea, just ignore it.

I know the go dep is not perfect, however by upgrading the dep I always get lots of changes from vendor folder. So far I never had any issue after removing vendor folder. reinstall_vendor_deps always works fine for me. If you want to keep it, you can ignore my changes.

Later if you want to upgrade to tendermint-18 you can count on me.

Thanks for your wonderful work on Burrow. It's well mangedand I love it.

PR #742, opened 2018-04-16
Better function naming

b00f:
Changes:
Rename `NewBlockChain` to `newBlockchain` and `LoadBlockchain` to `loadBlockchain`

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -100,7 +100,7 @@ type PersistedState struct {
 }
 
 func LoadOrNewBlockchain(db dbm.DB, genesisDoc *genesis.GenesisDoc,
-	logger *logging.Logger) (MutableBlockchain, error) {
+	logger *logging.Logger) (*blockchain, error) {
 
 	logger = logger.WithScope("LoadOrNewBlockchain")
 	logger.InfoMsg("Trying to load blockchain state from database",
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -100,12 +100,12 @@ type PersistedState struct {
 }
 
 func LoadOrNewBlockchain(db dbm.DB, genesisDoc *genesis.GenesisDoc,
-	logger *logging.Logger) (*blockchain, error) {
+	logger *logging.Logger) (MutableBlockchain, error) {
 
 	logger = logger.WithScope("LoadOrNewBlockchain")
 	logger.InfoMsg("Trying to load blockchain state from database",
 		"database_key", stateKey)
-	blockchain, err := LoadBlockchain(db)
+	blockchain, err := loadBlockchain(db)
 	if err != nil {
 		return nil, fmt.Errorf("error loading blockchain state from database: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


blockchain/blockchain.go
```diff
@@ -100,12 +100,12 @@ type PersistedState struct {
 }
 
 func LoadOrNewBlockchain(db dbm.DB, genesisDoc *genesis.GenesisDoc,
-	logger *logging.Logger) (*blockchain, error) {
+	logger *logging.Logger) (MutableBlockchain, error) {
```
silasdavis: it's not always correct but I think "accept interfaces, return structs" is a reasonable mantra here. An exception would be if we really can't trust the caller with a `*blockchain` - do you think this is the case here? Since `blockchain` is unexported it's really only for the benefit of this package - returning the struct makes it easier to test.

If anything I have been stepping back from my use of 'interfaces defined in same package as struct' in this case I think it is useful to define the `CommitBlock` function centrally but on reflection I'm questioning whether I need the `MutableBlockchain` at all...

blockchain/blockchain.go
```diff
@@ -100,12 +100,12 @@ type PersistedState struct {
...
 
 	logger = logger.WithScope("LoadOrNewBlockchain")
 	logger.InfoMsg("Trying to load blockchain state from database",
 		"database_key", stateKey)
-	blockchain, err := LoadBlockchain(db)
+	blockchain, err := loadBlockchain(db)
```
silasdavis: This is another issue where I am changed my own leaning (though open to debate on this) towards 'exported by default' rather than 'unexported by default'. In some languages it's common to hide as many functions as possible in service of encapsulation. While I agree that it's a good idea to limit what essentially becomes your 'public API' I think hiding (unexporting) a lot of functions can be an anti-pattern. For example using other libraries in a way perhaps the author didn't anticipate I have been bitten by needlessly unexported functions that are close relatives of ones that are exported. This feels like the case here.

It is unlikely than many people find our `blockchain` type that interesting, but as  a principle I'd rather export.

blockchain/blockchain.go
```diff
@@ -120,11 +120,11 @@ func LoadOrNewBlockchain(db dbm.DB, genesisDoc *genesis.GenesisDoc,
 	}
 
 	logger.InfoMsg("No existing blockchain state found in database, making new blockchain")
-	return NewBlockchain(db, genesisDoc), nil
+	return newBlockchain(db, genesisDoc), nil
 }
 
 // Pointer to blockchain state initialised from genesis
-func NewBlockchain(db dbm.DB, genesisDoc *genesis.GenesisDoc) *blockchain {
+func newBlockchain(db dbm.DB, genesisDoc *genesis.GenesisDoc) *blockchain {
```
silasdavis: Same as above - open to persuasion - but why hide this?

execution/execution_test.go
```diff
@@ -127,10 +127,16 @@ func makeUsers(n int) []acm.AddressableSigner {
 	}
 	return users
 }
+func newBlockchain(genesisDoc *genesis.GenesisDoc) bcm.MutableBlockchain {
```
silasdavis: This makes sense

blockchain/blockchain.go
b00f: In case we need to have Immutable-Blockchain, it might helps. I think your idea is better. Better to not make it so complicated. I will revert this change.

blockchain/blockchain.go
b00f: User wants to create a blockchain whenever he runs application. It can be created (first time) or loaded (other times). User shouldn't be involved in how we load/create blockchain. 
I vote for more encapsulation! however i's not a big change. I can also revert it.

blockchain/blockchain.go
silasdavis: :+1:

blockchain/blockchain.go
silasdavis: Okay, I'll take your vote, let's unexport this and `newBlockchain`

b00f: Thanks Silas.

PR #743, opened 2018-04-20
Add file and json logging instructions.

silasdavis:
With this change, running `burrow configure -l json,file,/tmp/foo.log` produces a burrow config containing:

```toml
[Logging]
  ExcludeTrace = false
  NonBlocking = false
  [Logging.RootSink]
    [Logging.RootSink.Output]
      OutputType = "file"
      Format = "json"
      Path = "/tmp/foo.log"
```


silasdavis:
I basically wrote a config to that is borderline unusable in toml... So I add to it as needed for debug. This was for looking at @b00f's testnet

b00f:
Logger makes life easier. It's good idea saving logs in file by default.

PR #744, opened 2018-04-20
Ensure storing zero value removes the key rather than panics

silasdavis:
`SetStorage` with a zero value is meant to indicate removal but in new IAVL would cause a panic. This resolves that by calling remove on underlying tree when a nil value is passed to execution state

See: https://github.com/hyperledger/burrow/blob/develop/vendor/github.com/tendermint/iavl/tree.go#L92

### DIFF
--- a/account/state/state.go
+++ b/account/state/state.go
@@ -33,7 +33,7 @@ type StorageGetter interface {
 }
 
 type StorageSetter interface {
-	// Store a 32-byte value at key for the account at address
+	// Store a 32-byte value at key for the account at address, setting to Zero256 removes the key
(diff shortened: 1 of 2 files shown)


account/state/state.go
```diff
@@ -33,7 +33,7 @@ type StorageGetter interface {
 }
 
 type StorageSetter interface {
-	// Store a 32-byte value at key for the account at address
+	// Store a 32-byte value at key for the account at address, setting to Zero256 removes the key
```
silasdavis: A comment along these lines exists in other places - I thought here too but apparently not. Adding it since it should be remarked.

PR #751, opened 2018-05-01
Implementing SHL, SHR and SAR opcodes

### LINKED ISSUE #705: [EVM] Support RETURNDATACOPY and RETURNDATASIZE opcodes
See: https://github.com/ethereum/EIPs/blob/dde2fe59532290f285a3ae79bd7347f0a7b69d6d/EIPS/eip-211.md

smblucker:
These changes implement bitwise shifting operations (shift left, shift right, shift arithmetic right) as described in EIP here: 

https://github.com/ethereum/EIPs/blob/e451b058521ba6ccd5d3205456f755b1d2d52bb8/EIPS/eip-145.md

This implementation resolves Issue here:

fixes #705


execution/evm/vm.go
```diff
@@ -493,6 +493,75 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
...
 			vm.Debugf(" => 0x%X\n", res)
 
+		case SHL: //0x1B
+			shift, popErr := stack.PopU64()
+
+			if popErr != nil {
+				return nil, firstErr(err, popErr)
+			}
+
+			sh := uint(shift)
```
seanyoung: If uint is 32 bit, then this statement truncates. Best to compare shift to 256, and cast to uint on the call to value.Lsh()

execution/evm/vm.go
```diff
@@ -493,6 +493,75 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
...
+			}
+
+		case SHR: //0x1C
+			shift, popErr := stack.PopU64()
+
+			if popErr != nil {
+				return nil, firstErr(err, popErr)
+			}
+
+			sh := uint(shift)
```
seanyoung: Same as before

execution/evm/vm.go
```diff
@@ -493,6 +493,75 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
...
+
+			value := stack.PopBigIntSigned()
+			zeroInt := big.NewInt(0)
+
+			if sh >= 256 && value.Cmp(zeroInt) >= 0 {
```
seanyoung: Rather than comparing to zeroInt, just call "value.Sign() >= 0"

execution/evm/vm.go
```diff
@@ -493,6 +493,75 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
...
+				reset := big.NewInt(0)
+				stack.PushBigInt(reset)
+				vm.Debugf(" %X << %X = %X\n", value, sh, reset)
+			}
+			if sh >= 256 && value.Cmp(zeroInt) < 0 {
```
seanyoung: value.Sign() < 0

execution/evm/vm.go
```diff
@@ -493,6 +493,75 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
...
+			}
+
+		case SAR: //0x1D
+			shift, popErr := stack.PopU64()
+
+			if popErr != nil {
+				return nil, firstErr(err, popErr)
+			}
+
+			sh := uint(shift)
```
seanyoung: remove cast here

smblucker:
Suggested changes were made. Thank you for the feedback.

execution/evm/vm.go
```diff
@@ -493,6 +493,68 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
 			stack.Push64(int64(res))
 			vm.Debugf(" => 0x%X\n", res)
 
+		case SHL: //0x1B
+			shift, popErr := stack.PopU64()
+
+			if popErr != nil {
+				return nil, firstErr(err, popErr)
```
seanyoung: According to EIP-145, any shift value > 256 should result in 0. This code returns an error if the shift value > 2^64.

execution/evm/vm.go
```diff
@@ -493,6 +493,68 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
...
+				stack.PushBigInt(shiftedValue)
+				vm.Debugf(" %X << %X = %X\n", value, shift, shiftedValue)
+			}
+
+		case SHR: //0x1C
+			shift, popErr := stack.PopU64()
+
+			if popErr != nil {
+				return nil, firstErr(err, popErr)
+			}
```
seanyoung: According to EIP-145, any shift value > 256 should result in 0. This code returns an error if the shift value > 2^64.

execution/evm/vm.go
```diff
@@ -493,6 +493,68 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
...
+				stack.PushBigInt(shiftedValue)
+				vm.Debugf(" %X << %X = %X\n", value, shift, shiftedValue)
+			}
+
+		case SAR: //0x1D
+			shift, popErr := stack.PopU64()
+
+			if popErr != nil {
+				return nil, firstErr(err, popErr)
+			}
```
seanyoung: According to EIP-145, any shift value > 256 should result in 0. This code returns an error if the shift value > 2^64.

execution/evm/vm.go
```diff
@@ -493,6 +493,66 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
 			stack.Push64(int64(res))
 			vm.Debugf(" => 0x%X\n", res)
 
+		case SHL: //0x1B
+			shift, popErr := stack.PopU64()
+
+			value := stack.PopBigInt()
+
+			if shift >= 256 || popErr != nil {
```
seanyoung: stack.PopU64() might fail because the stack itself underflowed (i.e. there is nothing on the stack, ErrDataStackUnderflow).
What we could do is define a new error in execution/evm/vm.go called something like ErrDataStackTypeInt64Overflow and set that error in vm.PopU64() and vm.Pop64(). Next we can compare the error here like
if shift >= 256 or err == ErrDataStackTypeInt64Overflow {
}
We would still need to return an error in the case that ErrDataStackUnderflow.

execution/evm/vm.go
```diff
@@ -493,6 +493,66 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
...
+				stack.PushBigInt(shiftedValue)
+				vm.Debugf(" %X << %X = %X\n", value, shift, shiftedValue)
+			}
+
+		case SHR: //0x1C
+			shift, popErr := stack.PopU64()
+
+			value := stack.PopBigInt()
+
+			if shift >= 256 || popErr != nil {
```
seanyoung: Same in this case

execution/evm/vm.go
```diff
@@ -493,6 +493,66 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
...
+			} else {
+				shiftedValue := value.Rsh(value, uint(shift))
+				stack.PushBigInt(shiftedValue)
+				vm.Debugf(" %X << %X = %X\n", value, shift, shiftedValue)
+			}
+
+		case SAR: //0x1D
+			shift, popErr := stack.PopU64()
+
+			value := stack.PopBigIntSigned()
```
seanyoung: Same here

silasdavis:
Thanks for this @smblucker - I was adding a review but thought it might be clearer as a PR into your branch which is here: https://github.com/smblucker/burrow/pull/1

There are 3 things it tweaks:

1. Simplify the logic of SAR
2. Use `%v` in format strings (`%X` is probably not what we want)
3. In tests its preferable to use `t.Logf` rather than `fmt.Printf` - you're following the convention that's already there, but I've changed that

Don't feel you have to merge my PR, just wanted to make some comments. Feel free to shoot back here or there.

execution/evm/vm.go
```diff
@@ -493,6 +493,52 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
 			stack.Push64(int64(res))
 			vm.Debugf(" => 0x%X\n", res)
 
+		case SHL: //0x1B
+			shift, value := stack.PopBigInt(), stack.PopBigInt()
```
silasdavis: `value` shadows the `value` argument to `call`. Can we change it to something else, I suggest just `x` as used in other arithmetic functions.

execution/evm/vm.go
```diff
@@ -493,6 +493,52 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
...
+
+			if shift.Cmp(Big256()) >= 0 {
+				reset := big.NewInt(0)
+				stack.PushBigInt(reset)
+				vm.Debugf(" %X << %X = %X\n", value, shift, reset)
+			} else {
+				shiftedValue := value.Lsh(value, uint(shift.Uint64()))
+				stack.PushBigInt(shiftedValue)
+				vm.Debugf(" %X << %X = %X\n", value, shift, shiftedValue)
```
silasdavis: We would like to show the integer value rather than the hex here like in other arithemetic methods so these `%X`s shoiuld be `%v`s

execution/evm/vm.go
```diff
@@ -493,6 +493,52 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
...
+				reset := big.NewInt(0)
+				stack.PushBigInt(reset)
+				vm.Debugf(" %X << %X = %X\n", value, shift, reset)
+			} else {
+				shiftedValue := value.Rsh(value, uint(shift.Uint64()))
+				stack.PushBigInt(shiftedValue)
+				vm.Debugf(" %X << %X = %X\n", value, shift, shiftedValue)
+			}
+
+		case SAR: //0x1D
```
silasdavis: I'd suggest simplifying what we have here to:

```go
			shift, x := stack.PopBigInt(), stack.PopBigIntSigned()

			if shift.Cmp(big.NewInt(256)) >= 0 {
				reset := big.NewInt(0)
				if x.Sign() < 0 {
					reset.SetInt64(-1)
				}
				stack.PushBigInt(reset)
				vm.Debugf(" %v << %v = %v\n", x, shift, reset)
			} else {
				shiftedValue := x.Rsh(x, uint(shift.Uint64()))
				stack.PushBigInt(shiftedValue)
				vm.Debugf(" %v << %v = %v\n", x, shift, shiftedValue)
}
```

execution/evm/vm_test.go
```diff
@@ -95,6 +95,476 @@ func TestVM(t *testing.T) {
...
+	var gas uint64 = 100000
+
+	//Shift left 0
+	bytecode := MustSplice(PUSH1, 0x01, PUSH1, 0x00, SHL, return1())
+	output, err := ourVm.Call(account1, account2, bytecode, []byte{}, 0, &gas)
+	value := []uint8([]byte{0x1})
+	expected := LeftPadBytes(value, 32)
+	assert.Equal(t, expected, output)
+
+	fmt.Printf("Result: %v == %v\n", output, expected)
```
silasdavis: I know we use `fmt.Printf` elsewhere but it would be preferable to use `t.Logf` that give more metadata and can be redirected to other writers.

smblucker:
@silasdavis I will make the changes

binary/integer.go
```diff
@@ -100,6 +101,10 @@ func S256(x *big.Int) *big.Int {
 	}
 }
 
+func Big256() *big.Int {
```
silasdavis: we should just export the one above - ideally we'd like to only allocate once which is achieved by a var initialisation

silasdavis:
I'm happy with this after the change to `big256` mentioned. Also the tests look good and comprehensive.

PR #754, opened 2018-05-03
Detect JSON from config file

b00f:
NOTE: Genesis file contains some JSON data (ex. crypto.PublicKey)
This patch is for detecting JSON format from TOML


silasdavis:
Thanks @b00f, I've fixed in #756 with https://github.com/silasdavis/burrow/blob/34ec0af729e184a79ccc55138e494f81b9cf568d/config/source/source.go#L44 - the intention is the same, but I'm inclined to keep the regex.

PR #756, opened 2018-05-04
Fix test issues, and reorganise CLI UX. Read validator address on startup.

silasdavis:
This PR:

- Fixes broken test that slipped through broken CircleCI notification on a previous PR
- Updates docs to point at roadmap and get keys and burrow running
- (Re)introduces `burrow serve` replacing `burrow` since current usage was not ideal and broke some uses of `Spec`
- Adds a smoke test for the mow.cli app
- Moves setting the `ValidatorAddress` from `burrow configure` to `burrow serve`, i.e. boot time. This allows us to use identical configuration for each validator in a pool in many cases and just specify the validator address. We can also set this via an environment variable

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,8 +105,9 @@ build_race:	check build_race_db build_race_client
 # build burrow
 .PHONY: build_db
 build_db: commit_hash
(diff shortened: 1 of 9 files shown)


compleatang:
LGTM

smblucker:
LGTM

PR #764, opened 2018-05-10
Introduce InputAccount on param to simplify JS

silasdavis:


dennismckinnon:
This all looks good to me

PR #767, opened 2018-05-11
returndatasize-returndatacopy implementation

smblucker:
These changes implement retrundatasize and returndatacopy; these changes attempt to resolve issue found here:

fixes https://github.com/hyperledger/burrow/issues/705

### DIFF
--- a/execution/evm/vm.go
+++ b/execution/evm/vm.go
@@ -42,6 +42,7 @@ var (
 	ErrMemoryOutOfBounds      = errors.New("Memory out of bounds")
 	ErrCodeOutOfBounds        = errors.New("Code out of bounds")
 	ErrInputOutOfBounds       = errors.New("Input out of bounds")
+	ErrReturnDataOutOfBounds  = errors.New("Return data out of bounds")
 	ErrCallStackOverflow      = errors.New("Call stack overflow")
 	ErrCallStackUnderflow     = errors.New("Call stack underflow")
 	ErrDataStackOverflow      = errors.New("Data stack overflow")
@@ -113,6 +114,7 @@ type VM struct {
 	nestedCallErrors []ErrNestedCall
 	publisher        event.Publisher
 	logger           *logging.Logger
+	returnData       []byte
 	debugOpcodes     bool
 	dumpTokens       bool
 }
@@ -666,6 +668,28 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
 			}
 			vm.Debugf(" => [%v, %v, %v] %X\n", memOff, codeOff, length, data)
 
+		case RETURNDATASIZE: // 0x3D
+			stack.Push64(int64(len(vm.returnData)))
+			vm.Debugf(" => %d\n", len(vm.returnData))
+
+		case RETURNDATACOPY: // 0x3E
+			memOff, outputOff, length := stack.PopBigInt(), stack.PopBigInt(), stack.PopBigInt()
+
+			end := new(big.Int).Add(outputOff, length)
+
+			if end.BitLen() > 64 || uint64(len(vm.returnData)) < end.Uint64() {
+				return nil, ErrReturnDataOutOfBounds
+			}
+
+			data := vm.returnData
+
+			memErr := memory.Write(memOff, data)
+			if memErr != nil {
+				vm.Debugf(" => Memory err: %s", memErr)
+				return nil, firstErr(err, ErrMemoryOutOfBounds)
+			}
+			vm.Debugf(" => [%v, %v, %v] %X\n", memOff, outputOff, length, data)
+
 		case BLOCKHASH: // 0x40
 			stack.Push(Zero256)
 			vm.Debugf(" => 0x%X (NOT SUPPORTED)\n", stack.Peek().Bytes())
@@ -834,6 +858,8 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
 			vm.Debugf(" => T:%X D:%X\n", topics, data)
 
 		case CREATE: // 0xF0
+			vm.returnData = nil
+
 			if !HasPermission(vm.stateWriter, callee, permission.CreateContract) {
 				return nil, ErrPermission{"create_contract"}
 			}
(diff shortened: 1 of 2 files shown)


execution/evm/vm.go
```diff
@@ -871,6 +913,8 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
 			} else {
 				newAccount.SetCode(ret) // Set the code (ret need not be copied as per Call contract)
 				stack.Push(newAccount.Address().Word256())
+				emptyBuffer := []byte{}
+				vm.setReturnDataBuffer(emptyBuffer)
```
silasdavis: When there is an exception After line 912 we need:

```go
vm.returnData = ret
```
since the EIP states:
> As an exception, CREATE and CREATE2 are considered to return the empty buffer in the success case and the failure data in the failure case

I also note that while I think you are following the spec here, geth and eth-cpp appear to always return what CREATE returns. Will mull this over.

However we should be clearing the buffer when we enter `CREATE` on line 879. Also there is no need to create an empty buffer, just use nil:

```go
vm.returnData = nil
```

And remove the clearing of the buffer where you currently have it. Either we will have set the buffer to ret on failure or it will be empty anyway.

execution/evm/vm.go
```diff
@@ -955,11 +999,13 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
 						return nil, firstErr(callErr, ErrUnknownAddress)
 					}
 					ret, callErr = vm.Call(callee, callee, acc.Code(), args, value, &gasLimit)
+					vm.setReturnDataBuffer(ret)
```
silasdavis: remove these in favour of setting once after this block. It's simpler and clearer and currently we are not capturing return data from native contracts, which we want to do.

execution/evm/vm.go
```diff
@@ -955,11 +999,13 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
...
 				} else if op == DELEGATECALL {
 					if acc == nil {
 						return nil, firstErr(callErr, ErrUnknownAddress)
 					}
 					ret, callErr = vm.DelegateCall(caller, callee, acc.Code(), args, value, &gasLimit)
+					vm.setReturnDataBuffer(ret)
```
silasdavis: remove

execution/evm/vm.go
```diff
@@ -971,6 +1017,7 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
 					// add account to the tx cache
 					vm.stateWriter.UpdateAccount(acc)
 					ret, callErr = vm.Call(callee, acc, acc.Code(), args, value, &gasLimit)
+					vm.setReturnDataBuffer(ret)
```
silasdavis: remove

execution/evm/vm.go
```diff
@@ -971,6 +1017,7 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
...
 				}
 			}
 			// In case any calls deeper in the stack (particularly SNatives) has altered either of two accounts to which
```
silasdavis: Add `vm.returnData = ret` here in place of removed setters above

We also need to clear the buffer on entry to `CALL*` on line 925 with `vm.returnData = nil` (a lot of the time this won't matter, but catches some edges cases on early return and matches the spec more clearly)

execution/evm/vm.go
```diff
@@ -666,6 +676,38 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
 			}
 			vm.Debugf(" => [%v, %v, %v] %X\n", memOff, codeOff, length, data)
 
+		case RETURNDATASIZE: // 0x3D
+			stack.Push64(int64(len(vm.getReturnDataBuffer())))
+			vm.Debugf(" => %d\n", len(vm.getReturnDataBuffer()))
+
+		case RETURNDATACOPY: // 0x3E
+			memOff := stack.PopBigInt()
+			outputOff, popErr := stack.Pop64()
```
silasdavis: Use `PopBigInt` here

execution/evm/vm.go
```diff
@@ -666,6 +676,38 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
...
+			if popErr != nil {
+				return nil, firstErr(err, popErr)
+			}
+			length, popErr := stack.Pop64()
```
silasdavis: Use `PopBigInt` here too

execution/evm/vm.go
```diff
@@ -853,6 +877,8 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
 				return nil, firstErr(err, ErrInsufficientBalance)
 			}
 
+			vm.returnData = nil
```
silasdavis: So I don't think this will matter in practice, at least not now, but I think strictly we should clear the `returnData` at the beginning of the frame, i.e. right after `CREATE:`. Doing this also makes us consistent with the other implementations.

execution/evm/vm.go
```diff
@@ -897,6 +923,8 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
 					return nil, firstErr(err, popErr)
 				}
 			}
+
+			vm.returnData = nil
```
silasdavis: here too, I just think it's marginally better immediately once we enter this case block.

silasdavis:
I've just noticed we are still missing this change:

silasdavis:
@smblucker we are missing this change I've just noticed;

When there is an exception After line 912 we need:

vm.returnData = ret

That is when there is an exception from a `CREATE` we need to assign the return to the return buffer

fulldecent:
In case anybody wanted to see. Here is a test case that proves it works:

```solidity
pragma solidity ^0.4.23;

contract ReturnsArray {
    function singleDigitPrimes() pure external returns (uint256[] memory) {
        uint256[] memory retval = new uint256[](4);
        retval[0] = 2;
        retval[1] = 3;
        retval[2] = 5;
        retval[3] = 7;
        return retval;
    }
}

contract ConsumesArray {
    uint256 public lowestSingleDigitPrime;
    
    constructor (ReturnsArray producer) public {
        uint256 x = producer.singleDigitPrimes()[0];
        lowestSingleDigitPrime = x;
    }
}
```

silasdavis:
Thanks @fulldecent! If you were so inclined we would welcome a pull request with your test case into: https://github.com/hyperledger/burrow/tree/develop/tests/jobs_fixtures now we have merged our `bos` tool into `burrow deploy`.

You can see how you might write it by looking at: https://github.com/hyperledger/burrow/blob/develop/tests/jobs_fixtures/app05-deploy_query_contracts_and_testing_rendering_solc_int_type/deploy.yaml.

Feel free to message me @silasdavis on https://chat.hyperledger.org/channel/burrow-contributors if you'd like some help putting it together.

fulldecent:
@silasdavis PR sent at https://github.com/hyperledger/burrow/pull/860

PR #771, opened 2018-05-14
Implementing shl, shr, and sar opcodes

smblucker:
These changes implement shl, shr, and sar opcodes.


silasdavis:
Thanks @smblucker !

PR #775, opened 2018-05-17
Update key.go

b00f:
This method generated wrong address from public_key.


b00f:
I found a bug on converting the public key to address. It's not very critical. You can merge this change later.
Better use crypto methods to use curve functionality.

silasdavis:
This was working around the fact Tendermint used go-wire to compute addresses and I guess avoiding depending on it. This change makes sense. Could you fix the tests?

I think you probably just need to `make fix` to fix goimports issues and signoff for DCO with `git commit -s --amend`

b00f:
@silasdavis. Please check it again. Thanks.

Issue #600, opened 2017-05-05
[EVM] Support REVERT opcode


silasdavis:
See: https://github.com/ethereum/EIPs/pull/206

seanyoung:
revert opcode merged, see https://github.com/hyperledger/burrow/pull/778

PR #778, opened 2018-05-23
Implementing revert opcode

smblucker:
These changes implement the revert opcode as referenced here:

https://github.com/hyperledger/burrow/issues/600

### DIFF
--- a/execution/evm/vm.go
+++ b/execution/evm/vm.go
@@ -105,7 +105,6 @@ type Params struct {
 }
 
 type VM struct {
-	stateWriter      state.Writer
 	memoryProvider   func() Memory
 	params           Params
 	origin           acm.Address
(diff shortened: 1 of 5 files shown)


execution/evm/vm.go
```diff
@@ -210,6 +210,8 @@ func (vm *VM) Call(caller, callee acm.MutableAccount, code, input []byte, value
 				return nil, fmt.Errorf("error transferring value %v %s (callee) -> %s (caller)",
 					value, callee, caller)
 			}
+		} else {
+			childCallState.Sync(callState)
```
silasdavis: I'm starting to feel `Sync` was a poor choice of name - that said I'm not sure i can think of a better one. Perhaps a comment here would be a nice addition:
```go
// Copy any state updates from child call frame into current call frame
```

execution/evm/vm.go
```diff
@@ -1145,16 +1150,17 @@ func (vm *VM) call(caller acm.Account, callee acm.MutableAccount, code, input []
...
+			callState.UpdateAccount(receiver)
+			callState.RemoveAccount(callee.Address())
 			vm.Debugf(" => (%X) %v\n", addr[:4], callee.Balance())
 			fallthrough
 
 		case STOP: // 0x00
 			return nil, nil
 
-		case STATICCALL, CREATE2:
+		case STATICCALL:
```
seanyoung: This PR is not changing anything wrt to CREATE2

smblucker: I made the changes including Silas' request for a comment above.

smblucker: Some tests are now not passing

PR #779, opened 2018-05-24
Updated GetData return value descriptor

TheAlienMind:
Used a JSON returned via RPC to understand the structure of returned object.


TheAlienMind:
Does the DCO check on merges as well? Seems to fail when I merge

PR #780, opened 2018-05-29
[WIP] Examine Finterra Burrow extensions

silasdavis:


LICENSE.md
```diff
@@ -1,176 +1,21 @@
...
-      or contributory patent infringement, then any patent licenses
-      granted to You under this License for that Work shall terminate
-      as of the date such litigation is filed.
-
-   4. Redistribution. You may reproduce and distribute copies of the
-      Work or Derivative Works thereof in any medium, with or without
-      modifications, and in Source or Object form, provided that You
-      meet the following conditions:
-
-      (a) You must give any other recipients of the Work or
```
silasdavis: I think you'll need to include the Apache 2.0 license - ideally stick with it

account/account.go
```diff
@@ -75,23 +75,25 @@ type Account interface {
...
-	SubtractFromBalance(amount uint64) (MutableAccount, error)
+	SubtractFromBalance(amount uint64) error
 	// Add amount to balance (will panic if amount plus balance is a uint64 overflow)
-	AddToBalance(amount uint64) (MutableAccount, error)
+	AddToBalance(amount uint64) error
 	// Set EVM byte code associated with account
-	SetCode(code []byte) MutableAccount
+	SetCode(code []byte)
 	// Increment Sequence number by 1 (capturing the current Sequence number as the index for any pending mutations)
-	IncSequence() MutableAccount
```
silasdavis: This enables a builder style syntax. Obviously a matter of taste, but better not to make inconsequential changes like this.

account/validator.go
```diff
@@ -1,93 +1,187 @@
...
+		stake:         stake,
+		bondingHeight: bondingHeight,
+		sequence:      0,
+	}
 }
 
-func AsConcreteValidator(validator Validator) *ConcreteValidator {
-	if validator == nil {
+func LoadValidator(bytes []byte) Validator {
+	u := map[string]string{}
```
silasdavis: I'm not that keen on this stringly typed way of encoding/decoding. Let's just define a struct that is a model of serialisation and call `json.Marshal` on that. Often we can use the same model of serialisation as model we use in-memory but if we can't I'd say just define one up front that clearly does the same thing on serialisation and deserialisation

silasdavis: Having said that the approach with a wrapper (providing the interface) and a concrete type providing the serialisation has it's downsides so I am open to alternatives

blockchain/blockchain.go
```diff
@@ -78,29 +83,25 @@ type tip struct {
 	lastBlockTime         time.Time
 	lastBlockHash         []byte
 	appHashAfterLastBlock []byte
+
+	validatorSet ValidatorSet
```
silasdavis: yep makes more sense

blockchain/blockchain.go
```diff
@@ -184,10 +204,27 @@ func (bc *blockchain) CommitBlock(blockTime time.Time, blockHash, appHash []byte
...
+		if err != nil {
+			return err
+		}
+
+		validatorSetJSON, err := bc.validatorSet.JSONBytes()
+		if err != nil {
+			return err
+		}
+
+		u := map[string]string{}
```
silasdavis: Yep as before I don't like this serialisation style much. I think the solution of explicitly defining the `PersistedState` is better, would rather that was reinstated.

blockchain/validator_set.go
```diff
@@ -0,0 +1,273 @@
...
+type ValidatorSet interface {
+	TotalPower() int
+	SetMaximumPower(maximumPower int)
+	MaximumPower() int
+	AdjustPower(height int64) error
+	Validators() []acm.Validator
+	SetLeavers() []acm.Validator
+	JoinToTheSet(validator acm.Validator) error
+	LeaveFromTheSet(validator acm.Validator) error
+	IsValidatorInSet(address acm.Address) bool
```
silasdavis: I suggest `Join`, `Leave`, and `Contains` are better more idiomatic names

blockchain/validator_set.go
```diff
@@ -0,0 +1,273 @@
...
+	return set
+}
+
+func (vs *validatorSet) TotalPower() int {
+	return len(vs.validators)
+}
+
+func (vs *validatorSet) SetMaximumPower(maximumPower int) {
+	if maximumPower > 90 {
+		maximumPower = 90
```
silasdavis: This looks like a stub, but would prefer constants for the magic numbers at least

blockchain/validator_set.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+
+	vs.maximumPower = maximumPower
+}
+
+func (vs *validatorSet) MaximumPower() int {
+	return vs.maximumPower
+}
+
+func (vs *validatorSet) AdjustPower(height int64) error {
```
silasdavis: This function could do with some explanatory comments

blockchain/validator_set.go
```diff
@@ -0,0 +1,273 @@
...
+	/// empty slice
+	vs.setLeavers = vs.setLeavers[:0]
+
+	dif := vs.TotalPower() - vs.maximumPower
+	if dif <= 0 {
+		return nil
+	}
+
+	limit := int(math.Floor(float64(vs.maximumPower*1/3))) - 1
```
silasdavis: Currently is not doing any float arithmetic, you mean `limit := int(math.Floor(float64(vs.maximumPower)/3)) - 1` but that itself can be simplified to just: `limit := vs.MaximumPower/3 - 1`, and it should be since in float arithmetic is not always deterministic.


However I am not convinced this is the value you mean, I think you want `limit := int(math.Ceil(float64(vs.maximumPower)/3)) - 1` since you want to limit `dif` to strictly less than 1/3 of maximum power (I think). So for example 10/3 = 3.33... so 3 is acceptable (< 1/3), whereas your formula would set it to 2 (floor(3.33) -1). Taking the ceiling would give for max power 10: ceil(10/3) - 1 = 3, for max power 9: ceil(9/3) -1 = 2 (which is what we want for strictly less than.

You can compute the ceiling with pure integer arithemtic with `limit := (vs.maximumPower + 2)/3` (the general integer formula is ceil(float64(n)/float64(d)) == (n + d - 1)/d for integers n,d (disregarding overflow))

blockchain/validator_set.go
```diff
@@ -0,0 +1,273 @@
...
+	Validators() []acm.Validator
+	SetLeavers() []acm.Validator
+	JoinToTheSet(validator acm.Validator) error
+	LeaveFromTheSet(validator acm.Validator) error
+	IsValidatorInSet(address acm.Address) bool
+	JSONBytes() ([]byte, error)
+}
+
+type validatorListProxy interface {
+	Validators(height int64) (*ctypes.ResultValidators, error)
```
silasdavis: Is this working around the fact we don't have a clean way to get the current validator set from ABCI? This is dubious and probably induces a race conditions. We really need to work against a set of validators and a particular commit height and have all of this happen while we hold the commit lock.

LICENSE.md
b00f: You are right. It was our mistake. I change the license to Apache 2.0.

account/account.go
b00f: No need to return the modified struct (here `MutableAccount` object) when the methods are called by reference (pointer).
 
First of all, it increases the performance; because no need to copy the whole object any time a modification is required. It just changes the internal object. It makes object more encapsulated and robust. 
Also it has better code understanding, `MutableAccount` is the instance of account that can be modified. In the other hand the, the `Account` is just a read-only instance and no modifications is allowed. Any object which needs to modify Account should have `MutableAccount` instance. This policy make the code more safe.
By just returning the error code we can check if the operation have failed or not. No need to have copy of the object again.

Here is the rules that I am following to check if a method should be called by pointer or not:
* if a method doesn't change any struct field (`const` method), it calls without reference (pointer)
* if a method is not a `const` method (it changes the struct fields), it calls by reference (pointer).

```
func (s Struct)ImmutableMethod() err {
...
}

func (s* Struct)MutableMethod() err {
...
}
```

account/account.go
```diff
@@ -271,9 +273,9 @@ func (caw concreteAccountWrapper) MarshalJSON() ([]byte, error) {
 }
 
 // Account mutation via MutableAccount interface
-var _ MutableAccount = concreteAccountWrapper{}
+var _ MutableAccount = &concreteAccountWrapper{}
 
-func (caw concreteAccountWrapper) SetPublicKey(pubKey PublicKey) MutableAccount {
+func (caw *concreteAccountWrapper) SetPublicKey(pubKey PublicKey) {
```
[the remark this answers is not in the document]
b00f: All methods in `Account` interface are calling without pointer,  however methods in `MutableAccount` interface are calling by pointer.
All methods in `MutableAccount` are changing a field in Account struct.

account/validator.go
b00f: I am not that keen to define a structure just only for serializing/serializing. You know the `go` is a naive programming language and it's not written to make life easier! What we have is to choose between two pains!

blockchain/validator_set.go
b00f: Thanks

blockchain/validator_set.go
b00f: Thanks.

blockchain/blockchain.go
b00f: Let we discuss more about it. Maybe there is third way. `PersistedState` is used just for serializing/deserializing. We can back to `PersistedState`, but don't you think this bring more complexity to the code.

blockchain/validator_set.go
b00f: You are right. I need to put more comments here. 
This algorithm is a little complicated and I tried to write it as simple as possible. If you look at the code, the main loop it's almost around 40 lines. The main idea behind this algorithm is comparing current validator set with previous sets. Any validator who is the current set but not in the previous set(s) has joined recently. I try to sort arrays before comparing them, so it helps to compare arrays easier and faster (the performance is `O(n log(n))`)... 
However, this piece of the code is very important in order to keep blockchain alive. Any mistakes here will cause blockchain goes dead.
About 1/3, The chance of changing 1/3 of the nodes in each round is almost zero. To make a blockchain safe and fast enough we figured out that a number between 60 and 80 validators is the best for validator set. (maximum power should be something between 60~80)
Take 60 as an example, It's almost impossible that 20 nodes nominating themselves to enter to the validator set in one round. Almost impossible! Impossible is possible, so we always make sure to not change more than 2/3 of nodes in each round. 
How about any possible attack? It's possible. Let me explain it a little more here.

Consider the VRF method is working fine and the byzantine party have no control over it. (we need to consult with some cryptographers to make sure about its safety. I believe we can make it safe with including some cure parameters into the proof. I am not worried about this part.)
If the VRF itself is safe, the only way to attack the blokchain is generating so many random pair keys and try to find suitable values for sending sortition transaction.
I try to prevent this attack by two changes:
1- The seed for the VRF in each round is almost random and unpredictable; It's previous block hash.
2- There is a gap between first bonding transaction and first sortition transaction.

I figured out that in a live network it might not possible for a node to send a sortition transaction just for entering into the next validation set. It might not be seen by every node at coming round. This transaction should be valid for more than one round, But it can not be valid forever. My suggestion is that the sortition transaction can be valid for just 10 rounds (I am very keen for changing it).
This is the only way an attacker can attack the blockchain. They can start generating valid sortition transaction by using lots of random generated key pairs. Since this sortiton transaction is valid for 10 rounds they can attack the blockchain in between.
The protocol predicts this attack, and we prevented it very simply.
Before a node starts sending a sortition transaction, it should bond some amount. Sortition transaction can only send 10 rounds after bonding transaction. After bonding transaction the node can not send any sortiton for first 10 rounds.

blockchain/validator_set.go
b00f: For two reason I defined the proxy interface here:
1- writing test independently
2- In future we might change the way we get validator list. It might be get directly from ABCI.

LICENSE.md
silasdavis: Great

account/account.go
silasdavis: > No need to return the modified struct (here MutableAccount object) when the methods are called by reference (pointer).

There is no strict 'need' but it allows a declarative builder syntax like:

```go
acc.IncSequence().AddToBalance(inputAmount).SubtractToBalance(outputAmount)
```
This is why the method returns the bash-in-place object. Now whether you like this builder syntax is a matter of taste. I happen to think it keeps some important operations on a single logically line and prevents accidentally interspersing operations that should be atomic with other calls (that may err for example).

More to the point though really is there is no reason to change this relating to your changes - so when we come to PR as a hope we can soon it either belongs in a separate PR (I know you didn't open this as PR of course :)! ) or should be reverted.

> First of all, it increases the performance; because no need to copy the whole object any time a modification is required. 

This is incorrect. It does not copy the object the return type is an interface which is a pointer type. The performance implications are therefore utterly negligible.

> It just changes the internal object. It makes object more encapsulated and robust.

All these methods just change the internal object. You'll have to explain why this makes them more encapsulated and robust - I don't see it.

> Also it has better code understanding, MutableAccount is the instance of account that can be modified. In the other hand the, the Account is just a read-only instance and no modifications is allowed. Any object which needs to modify Account should have MutableAccount instance. This policy make the code more safe.

This comment applies in its entirety to the current implementation - so I agree!

> By just returning the error code we can check if the operation have failed or not. No need to have copy of the object again.

You can still check the error code in the current implementation. As I have said there is not object copy.

account/account.go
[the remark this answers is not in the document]
silasdavis: `concreteAccountWrapper` just contains a pointer:

```go
type concreteAccountWrapper struct {
	*ConcreteAccount `json:"unwrap"`
}
```
Your change just makes us hold a pointer to a pointer - which buys us nothing. 

This looks like a case of you pattern matching on mutability - but in this case that pattern is wrong.

account/validator.go
silasdavis: Strongly disagree on this - I have had to maintain hand spun serialisation logic like this and it is a pain. It's too easy to change a string - and crucially map order is *guaranteed* to be non-deterministic which in general we do not want if we are ever hashing things.

It is much better to have a single declarative struct that defines how a stored object can be serialiased *and* deserialised in a single location - exported so that other packages are able to handle that serialised object should they need to. Not to mention the fact you can attach documentation to the struct.

Interspersing code with the busywork of serialising integers is a mess. Defining a struct 'just' for serialisation is a great use of a struct. Once again this is also a case of folliowing existing conventions in the codebase if you still don't find the above convincing.

blockchain/blockchain.go
silasdavis: It is possible to use an in-line anonymous struct for serialisation which would contain it to the method - however this misses out centralising our 'model of serialisation'. I think the implicit model of serialisation defined separately from the model of deserialisation is more complex.

A third way might be to separate our 'models of serialisation' from our 'models of computation' - it sounds to me like this kind of containment/separation might be what you are after. I wouldn't be against having serialisation models moved to the bottom of file, a separate file, or maybe even a sub-package if there were more of them (or if they were used elsewhere). This is comparable to what you might do with protobuf generated structs (also models of serialisation).  For which matter we may transition to at some point - though we would still have the structs.

blockchain/validator_set.go
silasdavis: That's would be good since it reflects your 1 unit of power for each validator design.

blockchain/validator_set.go
silasdavis: @seanyoung and I will be taking some time to study you design paper and look a this. 

I think you are right to identify what is effectively a stake grinding attack. This is how proof of work protocols can degenerate into proof of work. The usual mitigation for this is to slow down the rate at which stake can move around, which is effectively what I think you are saying if that you have to stake to a key pair before you can know the block hash which would allow you to throw computation at finding a valid sortition address/key.

I like this approach you are taking - but I do think it needs a careful pencil and paper analysis by a computer scientist/cryptographer.

blockchain/validator_set.go
silasdavis: Note you still have an error in the code `float64(vs.maximumPower*1/3)` does nothing because `vs.maximumPower*1/3` has already be integer divided.

You should just use `vs.maxmimPower/3 - 1` - it's simpler, faster, and more correct.

blockchain/validator_set.go
silasdavis: Yep I just wanted to check that was the case - interface is a good idea but the current implementation through the RPC is problematic and will lead to a race condition - we should definitely be getting this through the ABCI in the future.

account/validator.go
b00f: Serializing can be done without relying on Go functionality. We can make deterministic serilizable function, (it's our own implementation).

blockchain/validator_set.go
b00f: I also used rpc_core for checking sortition transaction. That part also should change. Thanks

b00f:
> Another thought I had was combining SendTx BondTx and UnbondTx into a single SendTx over as multi-asset balance.

I had the same idea, but since we need public-key we can't do it right now. Please check here: [https://github.com/tendermint/tendermint/issues/1625](tendermint-1625)

PR #782, opened 2018-06-04
Permission check

seanyoung:
Check that key files have the correct permission.


cmd/burrow/commands/keys.go
```diff
@@ -46,9 +48,26 @@ func Keys(output Output) func(cmd *cli.Cmd) {
 
 		cmd.Command("server", "run keys server", func(cmd *cli.Cmd) {
 			keysDir := cmd.StringOpt("dir", keys.DefaultKeysDir, "specify the location of the directory containing key files")
+			badPerm := cmd.BoolOpt("allow", false, "Allow unix key file permissions to be readable other than user")
```
silasdavis: feel like this should be a bit longer. `allow-bad-perm` or `ignore-perms` or something like that

keys/key_store_file.go
```diff
@@ -99,9 +100,15 @@ func IsValidKeyJson(j []byte) []byte {
 	return nil
 }
 
+func NewKeyStore(dir string, AllowBadFilePermissions bool, logger *logging.Logger) KeyStore {
+	return KeyStore{keysDirPath: dir, AllowBadFilePermissions: AllowBadFilePermissions, logger: logger.WithScope("KeyStore")}
```
silasdavis: This would be applicable for a `Scope` too but needs `ComponentKey` primarily. So could we have:

```go
logger: logger.With(structure.ComponentKey, "keys").WithScope("NewKeyStore")
```

Also let's make this initialisation multiline

keys/server_test.go
```diff
@@ -37,7 +38,7 @@ var (
 func init() {
 	failedCh := make(chan error)
 	go func() {
-		err := StartStandAloneServer(DefaultKeysDir, DefaultHost, TestPort)
+		err := StartStandAloneServer("test_scratch/"+DefaultKeysDir, DefaultHost, TestPort, false, logging.NewNoopLogger())
```
silasdavis: Oh let's also `os.RemoveAll(testDir)` at beginning of test, ala: https://github.com/hyperledger/burrow/blob/develop/core/integration/test_wrapper.go#L54-L56

PR #800, opened 2018-06-13
Refactoring Account object

b00f:
In this change list, the `Account` object is refactored and the interfaces are removed. 
Also `state_cache` has minor changes in order to keep a copy of account structure not the pointer. It is more safe because any further changes in account object  won't affect the state cache unless it gets update again.


silasdavis:
Thanks @b00f I think I like the direction this is going. I don't have time for a full review today so thought I'd dump a couple of things:

1. You'll need to make the integration tests pass `make test_integration` before we can merge. Before you do that make sure we've got the model right.

2. I like the KISS of `*Account`, however I do not want to lose the immutability when accounts are passed to objects that should not mutate them. I would like _some_ kind of read-only interface, for example we could do something like: https://github.com/hyperledger/burrow/blob/develop/blockchain/blockchain.go#L36-L44. I think we can dump `MutableAccount` though I prefer just using `*Account`

account/account.go
```diff
@@ -92,248 +52,155 @@ type ConcreteAccount struct {
...
 
-func DecodeConcrete(accBytes []byte) (*ConcreteAccount, error) {
-	ca := new(concreteAccountWrapper)
-	err := wire.ReadBinaryBytes(accBytes, ca)
+func (acc Account) MarshalJSON() ([]byte, error) {
+	if acc.validate() == false {
+		return nil, fmt.Errorf("Invalid account ")
+	}
+
+	return json.Marshal(acc.data)
```
silasdavis: yep this is the best way to do things

silasdavis:
Also needs a rebase now

b00f:
I merged it (not rebase)  with the latest changes.
Unfortunately I couldn't find out why the integration tests are failed. Could you please check it.

compleatang:
@b00f a bunch of things have recently been in flux at the integration level but are slowly stabilizing. Can you do a full rebase against current develop and push. It would be good to make sure you're changes are ran against an integration test we know is working rather than one that is in flux.

b00f:
I am working on refactoring test. So let me finish it then I will rebase this changes. The tests made me crazy.

PR #804, opened 2018-06-15
Simplify account structure and use struct for MutableAccount

silasdavis:
Based on the core idea of: https://github.com/hyperledger/burrow/pull/800 but more minimal and incremental

### DIFF
--- a/account/account.go
+++ b/account/account.go
@@ -55,32 +55,11 @@ type Account interface {
 	String() string
 }
 
-type MutableAccount interface {
-	Account
-	// Set public key (needed for lazy initialisation), should also set the dependent address
-	SetPublicKey(pubKey crypto.PublicKey) MutableAccount
-	// Subtract amount from account balance (will panic if amount is greater than balance)
-	SubtractFromBalance(amount uint64) (MutableAccount, error)
-	// Add amount to balance (will panic if amount plus balance is a uint64 overflow)
-	AddToBalance(amount uint64) (MutableAccount, error)
-	// Set EVM byte code associated with account
-	SetCode(code []byte) MutableAccount
-	// Increment Sequence number by 1 (capturing the current Sequence number as the index for any pending mutations)
-	IncSequence() MutableAccount
-	// Set the storage root hash
-	SetStorageRoot(storageRoot []byte) MutableAccount
-	// Set account permissions
-	SetPermissions(permissions ptypes.AccountPermissions) MutableAccount
-	// Get a pointer this account's AccountPermissions in order to mutate them
-	MutablePermissions() *ptypes.AccountPermissions
-	// Create a complete copy of this MutableAccount that is itself mutable
-	Copy() MutableAccount
+// MutableAccount structure
+type MutableAccount struct {
+	concreteAccount *ConcreteAccount
 }
 
-// -------------------------------------------------
-// ConcreteAccount
-
-// ConcreteAccount is the canonical serialisation and bash-in-place object for an Account
 type ConcreteAccount struct {
 	Address     crypto.Address
 	PublicKey   crypto.PublicKey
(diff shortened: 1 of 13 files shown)


b00f:
There are some changes here that I would like to discuss about them. It might help us to make the Burrow's project more flexible and robust.
But before that let me tell you why we decided to refactor `Account` object and how it's ended up to the PR #800.
In the Hyperledger-Burrow, Mutable and Immutable interfaces are introduced to work with account object and we didn't want to have different scheme for validator object. So we decided to follow the same approach.
In the other hand Account object has exported structures among the interfaces. It's something that I believe it's not a good practice and should be avoided. Interfaces in go are introduced to achieve polymorphism. We might use it to introduce some behaviors and functionality, but at the same time exporting the struct type will bring more complexity to the code. More important, by exporting struct type we allow the outside world to initialize the `account` object without having any control on it. 
This is why we make `account` struct unexported and introduced two constructors for initializing account object. This causes another problem, there is no way to using go packages for unmarshaling bytes to an interface! We solved it by implementing our own serialization methods. You have commented about it [here](https://github.com/hyperledger/burrow/pull/780#pullrequestreview-123992228). The serialization implementation is verbose. It could also be non-deterministic. 
We are in a trap! Having Mutable/Immutable interfaces is the root of the problem.

So I think to get out of the trap we should focus on this and answer this question. Is it a good approach having mutable and immutable interfaces for an object?

I have some critiques about it. First of all it's not a known design pattern (two interfaces for one structure). The restriction comes after make the code more complex.
In the other hand, object are created to be alive. The user of an object should know how to deal with the object. Also casting operation for switching between interfaces is not a good approach. 

So we decided to refactor the account object. This is a big change, but it will make codebase more readable and easy to understand. For an open source project it is very important having simple and easy to understand implementation. The code should be easy for everyone with less programming knowledge to understand and debug the code. 

In this PR the main problem still exists.

silasdavis:
> In the other hand Account object has exported structures among the interfaces. It's something that I believe it's not a good practice and should be avoided. 

Do you mean `ConcreteAccount`. Why do you think this should be avoided?

> More important, by exporting struct type we allow the outside world to initialize the account object without having any control on it.

What do we need to control? `ConcreteAccount` is a record type for which an exported type gives the cleanest initialisation syntax. Unless there is a clear safety reason then I think is unnecessary to hide the type, and hurts extension. In Go it is better to return structs and accept interfaces, having the struct public allows consumers to modify - where we want safety in execution we can use the interface or `MutableAccount` which has a `ConcreteAccount` as a private field.  

> We are in a trap! Having Mutable/Immutable interfaces is the root of the problem.
I do not see the argument for this. Also to be clear - this PR does remove the `MutableAccount` interface

I would like to draw attention to the following regarding this PR:
- My `Account` is still the read-only interfaces
- Your `Account` struct is my `MutableAccount` (now a struct with a private field)
- Your `accountData` is my `ConcreteAccount`
- Your `Account.data` is my `MutableAccount.concreteAccount` (note both private for when it matters - for mutations which we do want to control unlike initialisation)

Hopefully its clear from that how these changes are based on your account changes. However I have:
- Retained existing names which avoids renames in code for no good reason
- Exported `ConcreteAccount` - I disagree it is better not to export this serialisation model
- Kept a read-only `Account` interface

Your comment implies that there are still two interfaces - note that is factually incorrect. My assertion is that this PR solves all the problems your one does - indeed it is based on it - but isolates the changes and honours existing code and conventions.

If on second reading you think there are essential problems addressed in your PR not addressed here then please come back to me.

b00f:
I still believe  that `ConcreteAccount` should be unexported: `concreteAccount` and Account should be created via constructors.

seanyoung:
go doesn't have constructors (e.g. nothing will get called if you use new). Of course, you can create a function which returns a new struct ConcreteAccount (note struct, not class). We have
those functions already: NewConcreteAccount() and NewConcreteAccountFromSecret().

As you can see in those functions, nothing scary happens there so allowing users to access and
create their own ConcreteAccount is unlikely the be harmful. In fact, this is a common pattern in go.

b00f:
> As you can see in those functions, nothing scary happens there so allowing users to access and create their own ConcreteAccount is unlikely the be harmful. In fact, this is a common pattern in go.
You are right. This is a common pattern in Go. Sometimes just having different taste makes to think differently.

silasdavis:
merged into https://github.com/hyperledger/burrow/pull/804 as a dependency of getting gogo proto set up

PR #807, opened 2018-06-18
Solve a few glitches for integration tests on Mac and Linux environment

pietv:
This PR adds a few fixes for the burrow integration tests for them to run on a development machine:
* a check for dependent binaries that might not be present in a Docker container (`jq`, `xxd`, ...)
* The Darwin `base64` doesn't support the `-d` flag: change to `openssl base64 -d`.
* The LibreSSL `openssl base64 -d` command produces an empty result without the `-A` flag.
* Use the `trap EXIT` Bash command to kill the `burrow keys server` (occasionally the server keeps running after the integration tests failure).

### DIFF
--- a/keys/test.sh
+++ b/keys/test.sh
@@ -12,9 +12,26 @@ burrow_bin=${burrow_bin:-burrow}
 keys_dir=./keys/test_scratch/.keys
 
 echo "-----------------------------"
+echo "checking for dependent utilities"
+for UTILITY in jq xxd openssl; do
+        echo -n "... "
+        if ! command -v $UTILITY; then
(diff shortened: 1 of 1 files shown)


keys/test.sh
```diff
@@ -12,9 +12,26 @@ burrow_bin=${burrow_bin:-burrow}
...
+        exit 1
+fi
+
 echo "starting the server"
 $burrow_bin keys server --dir $keys_dir &
 keys_pid=$!
+function kill_burrow_keys {
+        kill -TERM $keys_pid
+}
+trap kill_burrow_keys EXIT
```
silasdavis: :+1:

PR #808, opened 2018-06-18
Implement ExecutionEvents service in GRPC

silasdavis:
This change implements persisting events to state binary serialised under a 16 byte key of `height index` in big endian format meaning that we can iterate events in order using IAVL's range query (and get a proof of the result if we want - though this is not currently exposed).

This core function added to state is:

```go
GetEvents(startBlock, endBlock uint64, queryable query.Queryable) (<-chan *events.Event, error)
```
and a GRPC based on this is provided by `ExecutionServer` that allows dynamic ranges (relative to latest block) to be specified as well as for events to be streamed continuously as they are generated by new blocks.

We sync events to our state tree before also broadcasting them over pubsub so existing functionality is still preserved.

This PR also:

- Reorganises protobuf and generation from makefile
- Fixes keys service using non-pointer receiver
- Fixes keys command checking keys options for nil rather than default values
- Adds `EventsService implementing with `Subscriptions`
- Adds `ExecutionEventsService` service state persisted events
- Further refines state handling and safety following through with some earlier changes to interfaces
- Improves read/write lock on state by maintaining a previous tree version for reading
- Provides an event endpoint with a query string
- Adds timeout to `SubscribeCallback` to prevent from blocking subscribers from halting pubsub
- Make multiple `NewBlock` event subscriptions safe


silasdavis:
The introduction of execution event wrapper means bosmarmot will need an upgrade before that integration test can pass

silasdavis:
If nowhere else, this needs a close read on execution/state.go and execution/execution.go

PR #814, opened 2018-06-20
Kubernetes secret

seanyoung:


cmd/burrow/commands/configure.go
```diff
@@ -16,8 +18,23 @@ import (
 	logging_config "github.com/hyperledger/burrow/logging/config"
 	"github.com/hyperledger/burrow/logging/config/presets"
 	"github.com/jawher/mow.cli"
+	"github.com/tendermint/go-amino"
+	tm_crypto "github.com/tendermint/go-crypto"
+	"github.com/tendermint/tendermint/p2p"
 )
 
+func processTemplate(pkg *deployment.Config, config *genesis.GenesisDoc, templateIn, templateOut string) error {
```
silasdavis: minor but can we put this at end, I like to see most important methods first

cmd/burrow/commands/configure.go
```diff
@@ -56,9 +77,11 @@ func Configure(output Output) func(cmd *cli.Cmd) {
 
 		chainNameOpt := cmd.StringOpt("n chain-name", "", "Default chain name")
 
-		cmd.Spec = "[--keys-url=<keys URL> | (--generate-keys=<secret keys files> [--keys-template=<text template for each key>])] " +
+		cmd.Spec = "[--keys-url=<keys URL> | --keysdir=<keys directory>] " +
+			"[--config-template-in=<text template> --config-template-out=<output file>]... " +
```
silasdavis: think `config-template-in` `config-out` better reflects semantics

cmd/burrow/commands/configure.go
seanyoung: Fixed

PR #820, opened 2018-06-26
Add healthcheck endpoint and prepare for release

silasdavis:
This PR primarily adds an endpoint suitable for implementing healthchecks, liveness probes, and readiness probes in rpc/TM as `last_block_info`. It reads data directly from memory and does not suffer from any possible database or lock contention like `status`. It also supports returning a 500 error if no blocks have been seen sufficiently recently.

When called bare it just returns the last block time, hash, and height: 
```javascript
http "localhost:46657/last_block_info"                
HTTP/1.1 200 OK
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: 
Access-Control-Expose-Headers: X-Server-Time
Content-Length: 196
Content-Type: application/json
Date: Tue, 26 Jun 2018 14:23:33 GMT
X-Server-Time: 1530023013

{
    "id": "",
    "jsonrpc": "2.0",
    "result": {
        "LastBlockHash": "81B11B917B1C8F9858127743A0F72B76474AA263",
        "LastBlockHeight": 3313,
        "LastBlockTime": "2018-06-26T15:23:32+01:00"
    }
}
```
When called with the parameter `block_within` which takes a `time.ParseDuration()` parsesable string then this endpoint will return with the same result as above if the last block committed has a `BlockTime` that falls within that duration counting back from now. If it does then the status is 200:
```javascript
http ":46657/last_block_info?block_within=5s"
HTTP/1.1 200 OK
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: 
Access-Control-Expose-Headers: X-Server-Time
Content-Length: 196
Content-Type: application/json
Date: Tue, 26 Jun 2018 14:21:47 GMT
X-Server-Time: 1530022907

{
    "id": "",
    "jsonrpc": "2.0",
    "result": {
        "LastBlockHash": "C7E38544D8DEBD01B04A1E69C1FB75BA4CB538B2",
        "LastBlockHeight": 3211,
        "LastBlockTime": "2018-06-26T15:21:46+01:00"
    }
}
```
If not then it returns an error with status code 500. The last block info is embedded in the error (since we cannot return an error and result under the JSONRPC 2.0 spec:
```javascript
http "localhost:46657/last_block_info?block_within=1ns"
HTTP/1.1 500 Internal Server Error
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: 
Access-Control-Expose-Headers: X-Server-Time
Content-Length: 350
Content-Type: application/json
Date: Tue, 26 Jun 2018 14:28:45 GMT
X-Server-Time: 1530023325

{
    "error": {
        "code": -32603,
        "data": "no block committed within the last 1ns (cutoff: 2018-06-26T15:28:45+01:00), last block info: {\"LastBlockHeight\":3611,\"LastBlockTime\":\"2018-06-26T15:28:44+01:00\",\"LastBlockHash\":\"197188B37FC8E5ABD0117B49F9C7180F35A43A4F\"}",
        "message": "Internal Error"
    },
    "id": "",
    "jsonrpc": "2.0"
}
```
We used a 1 nanosecond `block_within` interval in the above but the idea is the network operator can set the parameter to whatever they deem reasonable to probe either liveness or readiness. For simple liveness the no parameter version can be used.

This also prepares for the burrow 0.19.0 release and:
- Don't store Tx in state for every EventDataTx 
- Update changelog
- Remove logging that reveals private key for input account

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,50 @@
 # Hyperledger Burrow Changelog
+## Version 0.19.0
+This is a major (pre-1.0.0) release that brings upgrades, safety improvements, cloud configuration, and GRPC endpoints to Burrow.
+
+#### Breaking changes
+In addition to breaking changes associated with Tendermint (see their changelog):
+- State checkpointing logic has changed which has we load based on blockchain
+- Event format has changed over rpc/V0 see execution/events/ package
+- On-disk keys format has change from monax-keys to be more standard burrow keys
+- Address format has been changed (by Tendermint and we have followed suite) - conversion is possible but simpler to regenerated keys
+
+#### Features
+- Tendermint 0.20.0
+- Implemented EVM opcodes: REVERT, INVALID, SHL, SAR, SHR, RETURNDATACOPY, RETURNDATASIZE
+- Add config templating with burrow configure --config-template-in --config-out
+- Add config templates for kubernetes
+- Integrate monax-keys as internal (default) or standalone keys service, key gen exposed over CLI
+- Use GRPC for keys
+- Add GRPC service for Transactor and Events
+- Store ExecutionEvent by height and index in merkle tree state
+- Add historical query for all time with GetEvents
+- Add streaming GRPC service for ExecutionEvents with query language over tags
+- Add metadata to ExecutionEvents
+- Add BlockExplorer CLI for forensics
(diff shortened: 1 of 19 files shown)


compleatang:
LGTM

PR #829, opened 2018-07-08
Massively enhanced GRPC services including NameReg, transaction index, blocks service

silasdavis:
This PR has at its core an extension and refactoring of our GRPC services. There are three associated decisions that mean it has ended up as large as it is:

1. I have integrated gogoproto protobuf types tightly with many of our major types - this means everything eats the same dogfood internally as our RPC and gives us consistent models for most things
2. I have eliminate v0 completely and tm methods for events and broadcast - which entailed porting tests over to the new structure
3. I have stripped much of the event system complexity AND pulled in Tendermint's pubsub with a change to make it non-blocking

All of the RPC now runs off execution events called `BlockExecution` and `TxExecution` that mirror the Tendermint objects but contain the full execution trace as burrow sees it. These objects are meant to be self-describing and contain most data useful to querier. 

Rather than moving broken or redundant code I have tried to clean it up with the aim of making our breaking RPC changes all at once.


rpc/rpctransact/transact_server.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	return &TxEnvelope{
+		Envelope: txEnv,
+	}, nil
+}
+
+func (ts *transactServer) FormulateTx(ctx context.Context, param *PayloadParam) (*TxEnvelope, error) {
+	txEnv := param.Envelope(ts.transactor.Tip.ChainID())
+	if txEnv == nil {
+		return nil, fmt.Errorf("not payload provide to FormulateTx")
```
seanyoung: Not sure about this error, improve?

execution/events/tx.go
```diff
@@ -85,6 +91,31 @@ func SubscribeAccountOutputSendTx(ctx context.Context, subscribable event.Subscr
...
+	qry := query.NewBuilder().AndEquals(event.TxHashKey, hex.EncodeUpperToString(txHash))
+
+	subscriber, err := event.GenerateSubscriptionID()
+	if err != nil {
+		return err
+	}
+
+	// Acts as a sentinel for the final event
+	confirmationEventID := EventStringTxConfirmation(txHash)
+
```
seanyoung: Is it possible the tx will never make it to a block (consensus failure, another validator proposed block with tx for same address + same sequence number?) Do we leak stuff?

genesis/spec/template_account.go
```diff
@@ -10,19 +10,6 @@ import (
...
-	// Address  is convenient to have in file for reference, but otherwise ignored since derived from PublicKey
-	Address     *crypto.Address   `json:",omitempty" toml:",omitempty"`
-	NodeAddress *crypto.Address   `json:",omitempty" toml:",omitempty"`
-	PublicKey   *crypto.PublicKey `json:",omitempty" toml:",omitempty"`
-	Amount      *uint64           `json:",omitempty" toml:",omitempty"`
-	Power       *uint64           `json:",omitempty" toml:",omitempty"`
-	Permissions []string          `json:",omitempty" toml:",omitempty"`
-	Roles       []string          `json:",omitempty" toml:",omitempty"`
-}
-
```
seanyoung: tags are gone

rpc/rpctransact/transact_server.go
silasdavis: done

genesis/spec/template_account.go
silasdavis: done

PR #854, opened 2018-08-14
Move to new abi

seanyoung:
Note that the breaking change is what causes the bosmarmot test to fail.


project/history.go
```diff
@@ -30,7 +30,11 @@ func FullVersion() string {
 var History relic.ImmutableHistory = relic.NewHistory("Hyperledger Burrow", "https://github.com/hyperledger/burrow").
 	MustDeclareReleases(
 		"",
-		`Release our mempool signing lock once transactions have been CheckTx'd' to massively increase throughput, also support mempool signing for BroadcastTxAsync.`,
+		`Release our mempool signing lock once transactions have been CheckTx'd' to massively increase throughput, also support mempool signing for BroadcastTxAsync.
+
+#### Breaking changes
```
silasdavis: I know we used breaking changes in the past, but I thought we could just use:

\### Changed

As per: https://keepachangelog.com/en/1.0.0/

tests/jobs_fixtures/app33-evm_connection_edb_permissions_and_roles_layer/permissions.sol
```diff
@@ -9,62 +10,62 @@ pragma solidity >=0.4.0;
 * @dev Permissions permissions = Permissions(address(keccak256("Permissions")));
 */
 contract Permissions {
-  /**
-  * @notice Adds a role to an account
-  * @param _account account address
```
silasdavis: We had probably better put these back on reflection... It appears to be the style used by product team. It's to distinguish globals from returns. Shadowing is particularly bad in solidity because you tend to declare your returns outside of the body so it may be a little harder to notice.

PR #882, opened 2018-08-20
add proper docker labels per container label schema; s/monax/burrow/g on users

compleatang:


silasdavis:
incorporated into #883

PR #883, opened 2018-08-20
Sort out docker release

silasdavis:


Dockerfile
```diff
@@ -16,17 +16,17 @@ FROM alpine:3.8
 
 ARG REPO=/go/src/github.com/hyperledger/burrow
 
-ENV USER monax
-ENV MONAX_PATH /home/$USER/.monax
+ENV USER burrow
+ENV BURROW_PATH /home/$USER
 RUN addgroup -g 101 -S $USER && adduser -S -D -u 1000 $USER $USER
-WORKDIR $MONAX_PATH
+WORKDIR $ BURROW_PATH
```
compleatang: this won't work. can we just remove this file from the commit and use mine? it makes the same changes and adds a bit better hygeine to the image.

scripts/release.sh
```diff
@@ -2,10 +2,14 @@
 
 version_regex="^v[0-9]+\.[0-9]+\.[0-9]+$"
 
+set -e
 
 function release {
     notes="NOTES.md"
     echo "Building and releasing $tag..."
+    echo "Pushing docker image..."
+    docker login -u ${DOCKER_USER} -p ${DOCKER_PASS} && docker push ${DOCKER_REPO}
```
compleatang: docker best practice is...

```bash
echo ${DOCKER_PASS} | docker login --username ${DOCKER_USER} ${DOCKER_REPO} --password-stdin
```

Issue #857, opened 2018-08-15
[tests] refactor test contracts to remove warnings


compleatang:
Example:

```
Warning Generated during Contract Compilation
                                      Warning=ef50c3be35f6f68d9c4d26fbfeea0edd493bc8971de6e1ef2bd9ba87411bd1e3.sol:6:3: Warning: No visibility specified. Defaulting to "public". 
  function set(uint x) {
  ^ (Relevant source part starts here and spans across multiple lines).
ef50c3be35f6f68d9c4d26fbfeea0edd493bc8971de6e1ef2bd9ba87411bd1e3.sol:10:3: Warning: No visibility specified. Defaulting to "public". 
  function get() constant returns (uint retVal) {
  ^ (Relevant source part starts here and spans across multiple lines).
ef50c3be35f6f68d9c4d26fbfeea0edd493bc8971de6e1ef2bd9ba87411bd1e3.sol:17:2: Warning: No visibility specified. Defaulting to "public". 
	function create() returns (address GSAddr) {
 ^ (Relevant source part starts here and spans across multiple lines).
ef50c3be35f6f68d9c4d26fbfeea0edd493bc8971de6e1ef2bd9ba87411bd1e3.sol:22:2: Warning: No visibility specified. Defaulting to "public". 
	function getLast() returns (address GSAddr) {
 ^ (Relevant source part starts here and spans across multiple lines).
ef50c3be35f6f68d9c4d26fbfeea0edd493bc8971de6e1ef2bd9ba87411bd1e3.sol:22:2: Warning: Function state mutability can be restricted to view
	function getLast() returns (address GSAddr) {
 ^ (Relevant source part starts here and spans across multiple lines).
```

assertion still succeeds so NBD (for now)

compleatang:
migrated from https://github.com/monax/bosmarmot/issues/43

silasdavis:
Feel like we are some % done on this - @seanyoung ?

seanyoung:
Nearly all warnings are gone. solidity complains that some functions which contain "assembly mstore" should be declared pure, which is just wrong. Otherwise all done

PR #898, opened 2018-08-28
Fix solidity warnings, make tests use abi and add event/revert abi functionality

seanyoung:


seanyoung:
This fixes #857.

PR #912, opened 2018-09-14
Refactor state into refs and tree.

silasdavis:
This prepares the way for various state manipulations and introduces a top-level storage package that takes makes some steps towards Cosmos SDK compatibility. Changes included:

- `Prefix` and `KeyFormat` types for handling prefixed lexicographic keys consistently
- A `CacheDB` that wraps all changes within a block into a single DB transaction - i.e. Batch
- A separate 'tree' and 'refs' storage in state. The former describes those values that contribute to the state hash and the latter does not - this gives us more latitude in disabling, for example, block storage or running nodes that opt to store somewhat different derived data whilst agreeing the the core account state.
- Upgrade IAVL dependency to version including: https://github.com/tendermint/iavl/pull/107 

Some changes that may follow in a future PR:

- Content address account code for some deduplication
- Content address blocks to allow their hash to be stored in main tree for consistency (not sure about this one)
- Make block storage configurable
- Consider using a Trie implementation in various places including caches for some reduced string -> []byte copying.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -487,12 +487,12 @@
   version = "0.10.1"
 
 [[projects]]
-  digest = "1:76c7684f6ff9b92dd8f934ad853838e207ab4b7d68927f95f5ba785d8bea56d2"
+  digest = "1:95454de4bf99bf62e6e3b59724a712e954e1c8d9956e0bc098699179ad1168c6"
   name = "github.com/tendermint/iavl"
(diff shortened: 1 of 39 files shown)


seanyoung:
storage/cache_db.go line 45 contains a bug statement.
as @silasdavis is away let's merge anyway and I will fix up afterwards.

PR #920, opened 2018-09-25
Do not open connection for each meta job

seanyoung:


silasdavis:
yeah much nicer

PR #924, opened 2018-10-01
[consensus] Authorized only node sync / Peers filter support

### LINKED ISSUE #669: [consensus] Support authorised-only node sync
We need to implement the ability to lock down a network syncing with unauthorized peers. This was shipped in Tendermint 0.9 ish. The idea being that currently any node that can connect to the network will begin syncing and be able to pull the state. However, for a lot of use cases this is not desired and what is more desired is a network setup where validators only sync with validators/fullNodes that have been authorized.

Need to investigate how this functionality is provided by Tendermint and expose it through Burrow configuration.

phymbert:
Ability to lock down a network syncing with unauthorized peers. Currently any node that can connect to the network will begin syncing and be able to pull the state. However, for a lot of use cases this is not desired and what is more desired is a network setup where validators only sync with validators/fullNodes that have been authorized.

Designed to support static configuration or dynamic peers ids/addresses provided at peer handkshaking time.

fix #669

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -85,7 +85,7 @@ func (conf *BurrowConfig) Kernel(ctx context.Context) (*core.Kernel, error) {
 	}
 
 	return core.NewKernel(ctx, keyClient, privValidator, conf.GenesisDoc, conf.Tendermint.TendermintConfig(), conf.RPC,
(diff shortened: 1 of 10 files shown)


seanyoung:
That's a great feature, thank you

PR #926, opened 2018-10-04
[EVM] dynamically allocate stack depth, enforce call stack depth similarly

### LINKED ISSUE #386: [EVM] dynamically allocate stack depth
**problem** current max data stack capacity is hard-coded at 1024 words256; memory at 1MB (call depth is unlimited)

**solution** by default make data stack and memory unlimited; set optional limit in configuration file for embedded devices; enforce call stack depth similarly

phymbert:
fix #386 


execution/evm/stack.go
```diff
@@ -185,3 +189,29 @@ func Is64BitOverflow(word Word256) bool {
...
+		// Last attempt to grow
+		newCap = st.maxCapacity
+	}
+
+	// Grow
+	newData := make([]Word256, newCap)
+	copy(newData, st.data)
+	st.data = newData
+
+	return true
```
silasdavis: This whole function basically looks fine to me, but it occurs to me are doing basically the same thing here, but in a slightly different way: https://github.com/hyperledger/burrow/blob/develop/execution/evm/memory.go#L111-L142 

Thanks to Go and its type system we can't really reuse the implementation, but I wonder if we might want to borrow a few things from it:

1. It relies on the Go runtime to grow the backing array - this ends up being exponentially basically like you have done but has some optimisation for small sizes. I can't say it would make a dramatic difference, but I'd be inclined to piggy-back on Go's growth strategy which is more heavily optimised (and we get get at by using `append`).

2. There are some overflow checks there that we don't have here.

phymbert:
@silasdavis Yes I "reused" the same approach (+ I learnt a lot about slice and backed array), gr8t! please review

execution/evm/stack.go
```diff
@@ -21,23 +21,32 @@ import (
...
 	. "github.com/hyperledger/burrow/binary"
 	"github.com/hyperledger/burrow/execution/errors"
+	"math"
 )
 
+// Change the length of this zero array to tweak the size of the block of zeros
+// written to the backing slice at a time when it is grown. A larger number may
+// lead to fewer calls to append to achieve the desired capacity although it is
+// unlikely to make a lot of difference.
+var zeroWords []Word256 = make([]Word256, 32)
```
silasdavis: This ends up being 1 MiB which is quite a bit 'minimum chunk'. In memory.go we are just using 32 bytes. So `var zeroWord Word256` would be fine here. In fact we actually already have a constant you can use for this: https://github.com/hyperledger/burrow/blob/develop/binary/word256.go#L27. May as well import that and save the allocation.

Issue #933, opened 2018-10-08
[governance] Expose network setting to require automatically register validators in Registry


compleatang:
This is expected for the Agreements Network, but not for launch. The Network's validator pool should be relatively static. We should also have a defined space for registering these validators. Once a validator has agreed to the validators TOS then they should next register in nameReg their details. 

The data type envisioned is as follows:

```json
{
  "name": "validator/$VALIDATOR_ADDRESS",
  "value":
  {
    "moniker": "$VALIDATOR_MONIKER",
    "email": "$VALIDATOR_EMAIL"
  }
}
```

PR #936, opened 2018-10-09
[BIP] BIP2: Identify Transaction

phymbert:
Initially discussed on Hyperledger Chat, Burrow Contributors channel

Address #933


seanyoung:
I'm not sure IdentifyTx is descriptive enough. How about NodekeyTx?

silasdavis:
Had a quick pass through this - need to find some time to have a more thorough look. One meta question for now though is whether we should try and merge BIPs as 'fully formed', which wouldn't mean they could not longer be changed, or whether we should just merge based on whether the spirit of the thing is basically a good idea.

I can see advantages of both, on the one hand it seems like a lot of effort for the PRer to have to go through and make everyone else's edits so merging it would open up for subsequent tweaks as PRs. Then again I wouldn't want a half-finished or ill-conceived BIP to attract legitimacy just because it was in the repo if it ends up getting slightly abandoned.

A middle ground might be to merge proposed BIPs into their own branch - merging in them into master when they seem reasonably complete. We could use a special branch for this, or we could use develop.

Thoughts?

phymbert:
Hello @silasdavis , I suggest to try the last option, I prefer develop because it is the default branch and convenient to explore docs. We may highlight the status *draft* to inform readers that the community hasn't reached consensus yet on this BIP.

silasdavis:
Yes I like that idea, could you create a subfolder called draft and move this to that folder then I'll get this merged. We may want to drop the sequence number - but given there is no contention between BIPs now it doesn't really matter.

phymbert:
Hello @silasdavis , moved to draft subfolder done. Regarding BIP number id, I understand your point but I think we can face issue with links to assets folder - and this way it allows to keep some kind of history order between BIPs. Can we think about this later if we are facing duplicate number ? Another option is to use the BIP creation date as unique number, even I neither really like this way too.

silasdavis:
Agreed, let's figure it out later, thanks for your contribution.

PR #938, opened 2018-10-17
Empty deploy

seanyoung:


deploy/jobs/jobs_contracts.go
```diff
@@ -125,6 +125,9 @@ func DeployJob(deploy *def.Deploy, do *def.DeployArgs, client *def.Client, resp
 		if binaryResponse.Error != "" {
 			return "", fmt.Errorf("Something went wrong when you were trying to link your binaries: %v", binaryResponse.Error)
 		}
+		if binaryResponse.Binary == "" {
+			return "", fmt.Errorf("Contract binary code missing")
```
silasdavis: Can we add something like "Is your contract an abstract interface? You may not have unintentionally not implemented all abstracted inherited methods."

deploy/jobs/jobs_contracts.go
```diff
@@ -187,10 +193,14 @@ func DeployJob(deploy *def.Deploy, do *def.DeployArgs, client *def.Client, resp
 				if err != nil {
 					return "", err
 				}
+				deployedCount++
 				if strings.ToLower(response.Objectname) == strings.ToLower(strings.TrimSuffix(filepath.Base(deploy.Contract), filepath.Ext(filepath.Base(deploy.Contract)))) {
 					baseObj = result
 				}
 			}
+			if deployedCount == 0 {
+				return "", fmt.Errorf("Contract binary code missing")
```
silasdavis: maybe share error

silasdavis:
@seanyoung looks good other than making error messages a but more helpful

deploy/jobs/jobs_contracts.go
seanyoung: Good point, done

deploy/jobs/jobs_contracts.go
seanyoung: also true. done

PR #943, opened 2018-10-22
Test fixture for problems with forward proxy pattern

j-h-scheufen:
The implementation of a forward proxy pattern leads to `Error 5: Memory out of bounds` for functions returning a string (most likely this is a problem for all functions returning dynamic types).
This PR adds a test fixture to reproduce the problem.

Signed-off-by: Jan Hendrik Scheufen <j.h.scheufen@gmail.com>


silasdavis:
closing for #947

PR #967, opened 2018-11-16
 [evm] BLOCKHASH opcode

### LINKED ISSUE #909: [EVM] BLOCKHASH opcode unimplemented
- `burrow version`: master
- `go version`: 1.10
- what you expected to happen: To be able to use the `BLOCKHASH` opcode in Seth contracts
- what actually happened: The `BLOCKHASH` opcode is currently a no-op
- steps to reproduction: See vendored files in https://github.com/hyperledger/sawtooth-seth/pull/73

It would be useful to allow implementing the `BLOCKHASH` opcode in Burrow, as Seth provides this functionality (see https://github.com/hyperledger/sawtooth-seth/commit/b5475d45b2a4982fd83df1c7a572c284be65cf19 for where it was introduced to the vendored version of Burrow). The above-mentioned PR un-vendorizes Burrow, but runs into the issue of Burrow not allowing a way to call `GetBlockHash`.

The PR fixes this by overwriting bits of Burrow in the Dockerfile to introduce a `BlockHashGetter` interface used by `Reader`, and adding no-op implementations on relevant structs. This is a hack however, and should be added in Burrow proper. Is this something that could be added to Burrow?

phymbert:
close #909


core/kernel.go
```diff
@@ -112,6 +112,7 @@ func NewKernel(ctx context.Context, keyClient keys.KeyClient, privValidator tmTy
 	} else {
 		kern.State, err = execution.MakeGenesisState(stateDB, genesisDoc)
 	}
+	kern.Blockchain.BlockHashProvider = kern.State.GetBlockHash
```
silasdavis: Hm I'm not sure how I feel about patching it in like this. In the short term its nice and simple but it creates an ad hoc coupling between state and blockchain that doesn't pass through the standard constructor or other interfaces. I feel like it degrades our separate a bit. Let me reflect.

silasdavis:
Mostly looks good but can we:

1. Add `GetBlockHash` to `BlockchainInfo` interface
2. Add `GetBlockHash` to `evm/State` interface and pass `BlockchainInfo` as second argument to `evm.NewState` (state should capture the readers that the EVM needs now - we may want to allow for more info to be read from header)
3. Remove `GetBlockHash` from params (prefer to keep it static)

execution/state.go
```diff
@@ -348,7 +348,7 @@ func (s *State) GetTx(txHash []byte) (*exec.TxExecution, error) {
 }
 
 func (s *State) GetBlock(height uint64) (*exec.BlockExecution, error) {
-	bs := s.tree.Get(blockRefKeyFormat.Key(height))
+	bs := s.refs.Get(blockRefKeyFormat.Key(height))
```
silasdavis: Oh dear so this never worked!

silasdavis:
Sorry to delay on this, but every time I look at it I think it contains a single that the whole separation between blockchain and state is not done right. I do intend to get to that as part of working on the restore side of dump restore. I'd be inclined to merge it into that branch of work.

phymbert:
agreed, it is not done right... I will rebase onto develop the soonest. Meanwhile tell me if you have a better way to provide GetBlock to the evm.

phymbert:
Rebased ok, @silasdavis please confirm

silasdavis:
I am going to merge this - I don't think it makes sense to make it wait on rework - I'll take a look at that shortly though. All the basics are good.

PR #972, opened 2018-12-03
WIP: Fix checkpointing issue

silasdavis:
This currently just extends the kernel tests to reproduce the issue which is that in the case of rollback but the issue is with rewind and overwrite for which we need new IAVL


silasdavis:
subsumed by: #985

PR #975, opened 2018-12-07
Report stats on accounts with and without code

seanyoung:


acm/state/state.go
```diff
@@ -62,6 +71,7 @@ type Iterable interface {
 type IterableReader interface {
 	Iterable
 	Reader
+	AccountStatsGetter
```
silasdavis: We sure we want to pile this on IterableReader?

rpc/metrics/server.go
```diff
@@ -49,6 +49,8 @@ type Datum struct {
 	TxPerBlockBuckets   map[float64]float64
 	TotalTime           float64
 	TimePerBlockBuckets map[float64]float64
+	AccountsWithCode    uint64
+	AccountsWithoutCode uint64
```
compleatang: Can't add a comment on the line but we need to add the appropriate metrics into `func AddMetrics()` with something like:

```go
burrowMetrics["Contracts"] = prometheus.NewDesc(
	prometheus.BuildFQName("burrow", "accounts", "contracts"),
	"Current contracts on the chain",
	[]string{"chain_id", "moniker"}, nil,
)
burrowMetrics["Users"] = prometheus.NewDesc(
	prometheus.BuildFQName("burrow", "accounts", "users"),
	"Current users on the chain",
	[]string{"chain_id", "moniker"}, nil,
)
```

and also something in `func (e *Exporter) Collect(ch chan<- prometheus.Metric)` that looks like this:

```go
ch <- prometheus.MustNewConstMetric(
	e.burrowMetrics["Contracts"],
	prometheus.GaugeValue,
	e.datum.AccountsWithCode,
	e.chainID,
	e.validatorMoniker,
)
ch <- prometheus.MustNewConstMetric(
	e.burrowMetrics["Users"],
	prometheus.GaugeValue,
	e.datum.AccountsWithoutCode,
	e.chainID,
	e.validatorMoniker,
)
```

PR #988, opened 2018-12-19
Ensure tree version is synchronised with height

silasdavis:
Since the version of the IAVL tree is written into the node structure with every write and contribute to hash we much ensure that it is consistent between reloads.

Currently if no writes are made between successive commits we have:

```
AppHash_A -> CommitID {Version: 100 } written at Height 99
AppHash_A -> CommitID {Version: 101} written at Height 100
```
But suppose we attempt to restore starting from height 99. Now we end up starting with version 101 - which we are _meant_ to be overwriting. This will lead to AppHash mismatch on the next write.

To make things less confusing this also synchronises height and tree version (and checks this each commit).

With IAVL 0.12.0 we can make the checkpointing/rollback work as really intended with `LoadVersionForOverwriting`

### DIFF
--- a/execution/state.go
+++ b/execution/state.go
@@ -54,15 +54,18 @@ var (
 	storageKeyFormat  = storage.NewMustKeyFormat("s", crypto.AddressLength, binary.Word256Length)
 	nameKeyFormat     = storage.NewMustKeyFormat("n", storage.VariadicSegmentLength)
 	proposalKeyFormat = storage.NewMustKeyFormat("p", sha256.Size)
+
 	// Keys that reference references
 	blockRefKeyFormat = storage.NewMustKeyFormat("b", uint64Length)
 	txRefKeyFormat    = storage.NewMustKeyFormat("t", uint64Length, uint64Length)
-	// Reference keys
+
+	// Reference keys (that do not contribute to state hash)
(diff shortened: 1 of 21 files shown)


execution/state.go
```diff
@@ -54,15 +54,18 @@ var (
...
+
+	// Reference keys (that do not contribute to state hash)
 	// TODO: implement content-addressing of code and optionally blocks (to allow reference to block to be stored in state tree)
 	//codeKeyFormat   = storage.NewMustKeyFormat("c", sha256.Size)
 	//blockKeyFormat  = storage.NewMustKeyFormat("b", sha256.Size)
-	txKeyFormat     = storage.NewMustKeyFormat("b", tmhash.Size)
-	commitKeyFormat = storage.NewMustKeyFormat("x", tmhash.Size)
+	txKeyFormat = storage.NewMustKeyFormat("b", tmhash.Size)
+	// Binding between apphash and version stto
+	commitKeyFormat = storage.NewMustKeyFormat("v", uint64Length)
```
seanyoung: commitKeyFormat changed, this is used during startup to load the correct apphash. How can this be compatible?

silasdavis: yep... it won't be, this will need a fresh sync, would have been worth mentioning

PR #991, opened 2019-01-02
WIP: Trie implementation

silasdavis:
This is a prototype implementation of a space-efficient sparse trie structure based on compressed branching and a sparse child array at each branch.

As a trie it supports range queries and is lexicographically sorted. It also makes restricting to or iterating over a prefix straightforward. This has applications for our various caching layers that are currently map based and for state where if we extended it to form a merkle tree it would be easy to maintain hashes for any prefix (making it possible to have finer grain hashes for accounts and so on in a way not possible with IAVL).


silasdavis:
Gave this its own repo for development as a side-project: https://github.com/monax/trieste

PR #1000, opened 2019-01-24
Introduce forest of trees for state

silasdavis:
This manages a collection of trees rather than a single monolith tree.

With IAVL this will allow us in the future to: 

- Independently version sub-trees
- Obtain storage, proposal, name, or account sub-state hashes


storage/mutable_forest.go
```diff
@@ -0,0 +1,159 @@
...
+//   tree, err := forest.Writer("names")
+//   tree.Set("Cora", "unspecified")
+//
+// If there is no tree currently stored at the prefix passed then it will be created when the forest is saved:
+//
+//  hash, version, err := forest.Save()
+//
+// where the global version for the forest is returned.
+
+type MutableForest struct {
```
phymbert: Great description and implementation

bcm/blockchain_test.go
```diff
@@ -1,53 +1,43 @@
...
-	// Should have exponentially decayed to 0
-	assertZero(t, flow)
-	assertZero(t, bc.validatorCache.Power(id1.GetAddress()))
-}
-
-// Since we have -0 and 0 with big.Int due to its representation with a neg flag
-func assertZero(t testing.TB, i *big.Int) {
-	assert.True(t, big0.Cmp(i) == 0, "expected 0 but got %v", i)
-}
+//func TestBlockchain_Encode(t *testing.T) {
```
phymbert: could it be restored in this PR?

storage/mutable_forest.go
silasdavis: Glad you agree... went through a number of indecisions...

bcm/blockchain_test.go
silasdavis: I was going to port this, but I decided not to in the end - bit of weird development test and I wasn't convinced of value

PR #1003, opened 2019-01-31
BlockHash improvements

silasdavis:
Based off: https://github.com/hyperledger/burrow/pull/1000 so make sure that is merged before reviewing this.

This PR:
- Provides `Blockchain` with access to Tendermint's `Blockstore`. I anticipate us needing this if we move to not duplicating tendermint information in our storage anyway and feels a bit cleaner than what we have.
- Fixes that current implementation uses `AppHash` - it is a little arguably but probably it should be using the tendermint block hash proper
- Fixes off-by-one with 256 lookback on `BLOCKHASH` opcode


execution/evm/vm.go
```diff
@@ -617,8 +618,9 @@ func (vm *VM) execute(callState Interface, eventSink EventSink, caller, callee c
 			if blockNumber > vm.params.BlockHeight {
```
seanyoung: The "current" block is not committed so has no hash. Should this be "blockNumber >= vm.params.BlockHeight"

silasdavis: This is actually the last block so I think we are good: https://github.com/hyperledger/burrow/blob/develop/execution/contexts/call_context.go#L135

?

silasdavis: I think I agree with you remark - we should be passing the height of the block being applied

PR #1006, opened 2019-02-04
Change unit of aggregation in state to BlockEvent

silasdavis:
With the inclusion of dump/restore (in its current incarnation) we have the possibility of a very large zeroth block. Each time we dump/restore a chain we import all EVM log events into a zeroth pseudo-block. So this block can grow without bounds.

However GRPC has a relatively small message size limit of 4MB (https://github.com/grpc/grpc/issues/7927). The streaming RPC is designed to be used for large messages by breaking them into multiple parts. We have a natural batching with block and transaction but currently we store entire blocks in state. We would like to be able to naturally stream blocks in terms of the their transactions, however just streaming transactions makes it difficult to know when you have received the last transaction in a block. We work round this by introducing `BlockEvent` that can be either `BeginBlock`, `TxExecution`, or `EndBlock` (similar to the tendermint ABCI types). The block messages serve as delimiters.

This PR changes our fundamental execution event state to a stream of `BlockEvents`. Some remarks on this approach:

-  We can pull out a single transaction without decoding an entire (potentially very large) `BlockExecution`
- It relegates blocks (from the point of view of `State`) to being secondary to execution events (which kind of feels right given the separation we have between 'our state' and 'tendermint's state'. The aim is that state is only faintly aware it is running on a blockchain (whether this really makes sense is up for debate - but it does have some advantages for things like checkpointing and restore)
- We could go further and wrap `TxExecution` so we have a 'pure' event stream - for various reasons this seemed like going too far (would affect a lot of code and require a lot of reassembly). I decided in the end that `TxExecution` was a good unit to work with (indeed it is a transaction).
- However since block boundaries are our synchronisation markers, so when consuming a stream of transactions you often want to know when you have reached the end of a block (and code is nicer if you can still do this with a single stream).
- Consumers can still consume `BlockExecutions` with the function `rpcevents.ConsumeBlockExecutions` hiding the details of their reconstruction - so they just benefit from the streaming transport.

This PR also includes:
- KeyFormatStore construct to make sure we do not have colliding prefixes in state and improve readability
- Slight change to Vent - it chose to remain in an invite loop when receiving EOF (on non-streaming bounds) which did not seem like the right thing to do.


protobuf/exec.proto
```diff
@@ -19,10 +19,28 @@ option (gogoproto.sizer_all) = true;
 option (gogoproto.goproto_registration) = true;
 option (gogoproto.messagename_all) = true;
 
+message BlockEvent {
```
seanyoung: Now we have BlockExecution and BlockEvent which both contain TxExecutions.

How about naming BlockEvent -> BlockExecutionStream or StreamingBlockExecution?

PR #1010, opened 2019-02-06
Dump restore fixes

seanyoung:
Various fixes to dump/restore:
 - dump contains chain id and blocktime (populate Origin)
 - benchmask for loading mock restore file
 - require AppHash on restore
 - works with t5 dump


execution/state/state.go
```diff
@@ -130,7 +130,7 @@ func NewState(db dbm.DB) *State {
 }
 
 // Make genesis state from GenesisDoc and save to DB
-func MakeGenesisState(db dbm.DB, genesisDoc *genesis.GenesisDoc) (*State, error) {
+func MakeGenesisState(db dbm.DB, genesisDoc *genesis.GenesisDoc, restore DumpReader) (*State, error) {
```
silasdavis: Can we split this into 3 functions: one private one with sig as currently, one as previously, and one with current sig that is public called `MakeGenesisStateFromDump` or similar.

execution/state/dump_test.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+type MockDumpReader struct {
+	accounts int
+	storage  int
+	names    int
+	events   int
+}
+
+func (m *MockDumpReader) Next() (*dump.Dump, error) {
```
silasdavis: sweet

execution/state/state.go
seanyoung: If split it in another way, please let me know what you think

PR #1013, opened 2019-02-07
Vent triggers and projections refactor

silasdavis:
The primary purpose of this PR was to add the ability to specify event notification triggers using postgres' `LISTEN/NOTIFY` functionality. This is a core part of providing synchronised vent/burrow.js access and can also be used to build a resilient external event system.

In particular the log table now emits a `{"_height": "<the block height as string>"}` payload on the `height` channel for chain synchronisation purposes - wait for the height in a `TxExecution` and you know the database will be caught up with chain (for exact synchronisation with fields you can set a notification on a particular event table and subset of fields).

By specifying `Notify` channels as an array in the `EventClass` that table will register a row trigger that fires an event on each channel that is referenced in that event class/table spec. The payload sent by `NOTIFY` will be the JSON serialised collection of key values from `FieldMappings` for each mapping in which a particular channel is referenced.

```
  {
    "TableName": "EventTest",
    "Filter": "EventType = 'LogEvent'",
    "DeleteMarkerField": "__DELETE__",
    "FieldMappings": [
      {
        "Field": "key",
        "ColumnName": "testkey",
        "Type": "bytes32",
        "Primary": true,
        "Notify": ["keyed_meta"]
      },
      {
        "Field": "name",
        "ColumnName": "testname",
        "Type": "bytes32",
        "Primary": true,
        "BytesToString": true,
        "Notify": ["meta", "keyed_meta"]
      },
      {
        "Field": "description",
        "ColumnName": "testdescription",
        "Type": "bytes32",
        "Primary": false,
        "Notify": ["meta", "keyed_meta"]
      }
    ]
  },
```

In adding this functionality it was clear there were a number of problems with the structure being used to represent mappings, so included in this PR are various changes (including breaking changes that I will document with a subsequent commit), including:

- Case sensitive column names
- Clear separation between event field names and the columns to which they map
- Deletion marker made less busy just expecting a named field
- `burrow vent schema` now generates a JSON schema to check the event spec against
- EventSpecs can be merged together provided they agree on table definitions exactly
- Height is handled as a uint64 rather than a string (though still stored as varchar - we may want to use numeric but the way things work it doesn't hurt us as it is)
- A number of renames and restructuring to make the code more logical and easier to understand (e.g. `s/Parser/Projection` - parser was not a parser)


rpc/rpcevents/execution_events_server.go
```diff
@@ -193,14 +193,15 @@ func (ees *executionEventsServer) subscribeBlockExecution(ctx context.Context, c
...
 }
 
-func (ees *executionEventsServer) iterateBlockEvents(start, end uint64, consumer func(*exec.StreamEvent) error) (lastHeightSeen uint64, err error) {
-	err = ees.eventsProvider.IterateStreamEvents(start, end, func(blockEvent *exec.StreamEvent) error {
-		if blockEvent.TxExecution == nil {
-			// nil safe
-			lastHeightSeen = blockEvent.GetEndBlock().GetHeight()
+func (ees *executionEventsServer) iterateBlockEvents(start, end uint64, consumer func(*exec.StreamEvent) error) (uint64, error) {
+	// Assume that we have seen the previous block before start to have ended up here
+	lastHeightSeen := start - 1
```
seanyoung: That looks like a an underflow waiting to happen, i.e. if we iterate from block 0 then lastHeightSeen will be -1 and the leveldb reader does not deal with negatives well.

silasdavis: You are right, I have added a comment that notes this is true but we get away with it in this case.

PR #1020, opened 2019-02-17
[EVM] Support EXTCODEHASH

### LINKED ISSUE #996: Support EXTCODEHASH
See: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1052.md

phymbert:
Fixes #996


gregdhill:
LGTM

phymbert:
Thanks @gregdhill , wait before merging, I am trying to add an integration test. Also @silasdavis should we check as in `EXTCODECOPY` if the address references a native contract ?

phymbert:
Well, it is not possible to add intergation test at the moment, `extcodehash` is marked as `C`

> Constantinople is still in planning and all instructions marked as such will result in an invalid instruction exception

OK to merge this first implementation as is

seanyoung:
This might expose a problem in our implementation of the evm. When we deploy code via CallTx, the constructor arguments are concatenated to the bytecode. So, in effect, a contract with a string constructor foo("a") will have a different hash than foo("b").

PR #1027, opened 2019-02-20
Testcase for EXTCODEHASH

seanyoung:


seanyoung:
Looks like it depends on solidity 0.5 which has not been merged yet.

phymbert:
Yes, I mentioned it in the extcodehash PR. And even in 0.5 it looks extcodehash is marked as Constantinople, thus it will be generated as invalid, isnt?

PR #1029, opened 2019-02-22
Upgrade to Tendermint 0.30.1

silasdavis:
And upgrade other deps


deploy/loader/load_package_test.go
```diff
@@ -19,7 +19,7 @@ func TestUnmarshal(t *testing.T) {
     source: foo
     target: bar
     permissions: [foo, bar]
-    roles: []
+    roles: ["foo"]
```
silasdavis: `ni`l vs `[]` normalisation issue - this was easiest work around for tests

PR #1034, opened 2019-02-26
Dump

seanyoung:


execution/state/accounts.go
```diff
@@ -126,7 +126,7 @@ func (ws *writeState) SetStorage(address crypto.Address, key, value binary.Word2
 }
 
 func (s *ReadState) IterateStorage(address crypto.Address, consumer func(key, value binary.Word256) error) error {
-	keyFormat := keys.Storage
+	keyFormat := keys.Storage.Fix(address)
```
silasdavis: Oh, whoops. Does this not have a test?

PR #1039, opened 2019-02-27
Use consistent Signature message type

silasdavis:
Signatures in `Signatory` were being encoded as protobuf bytes of signature message rather than signature message itself via the use of gogoproto custom type dating from a time when `type Signature []byte`...


cmd/burrow/commands/keys.go
```diff
@@ -286,12 +286,22 @@ func Keys(output Output) func(cmd *cli.Cmd) {
 				if err != nil {
 					output.Fatalf("failed to hex decode message: %v", err)
 				}
+				curveType, err := crypto.CurveTypeFromString(*curveTypeOpt)
+				if err != nil {
+					output.Fatalf("invalid curve type: %v", err)
+				}
 
-				signature, err := hex.DecodeString(*sig)
+				fmt.Println(len(*sig))
```
seanyoung: debug printf?

PR #1069, opened 2019-03-12
Downgrade to Tendermint 0.29.1

silasdavis:


silasdavis:
Not Tendermint's fault...

PR #1075, opened 2019-03-19
Update multiple validators docs with new --test-burrow-configs

### LINKED ISSUE #1055: Update multiple-validators.md and enhance multi-node local setup
Current docs on multiple validators does not quite work, see: https://github.com/hyperledger/burrow/issues/1050

Also could it be streamlined? In particular `burrow configure` can generate node keys ahead of time. We should be able support this kind of try-out case without a lot of manual/scripted config text manipulation.

seanyoung:
Fixes #1055


phymbert:
LGTM very simpler starter, great!

PR #1123, opened 2019-05-30
Store TxExecution as single entry per block

seanyoung:


protobuf/exec.proto
```diff
@@ -81,6 +73,13 @@ message BlockExecution {
     repeated TxExecution TxExecutions = 3;
 }
 
+message TxExecutionKey {
+    // The block height
+    uint64 Height = 1;
+    // The offset of the TxExecution
+    uint64 offset = 2;
```
silasdavis: Capitalise...

protobuf/exec.proto
```diff
@@ -81,6 +73,13 @@ message BlockExecution {
     repeated TxExecution TxExecutions = 3;
 }
 
+message TxExecutionKey {
+    // The block height
+    uint64 Height = 1;
+    // The offset of the TxExecution
```
silasdavis: In bytes? If so maybe `OffsetInBytes` `OffsetBytes`?

execution/exec/block_execution.go
```diff
@@ -38,16 +29,25 @@ func (be *BlockExecution) StreamEvents() StreamEvents {
 	})
 }
 
-func (be *BlockExecution) Encode() ([]byte, error) {
-	return cdc.MarshalBinaryBare(be)
+func DecodeTxExecutionKey(bs []byte) (*TxExecutionKey, error) {
```
silasdavis: this sounds like it should be in tx_execution.go

execution/state/events.go
```diff
@@ -1,64 +1,83 @@
...
-			txHashTree.Set(keys.TxHash.KeyNoPrefix(ev.BeginTx.TxHeader.TxHash), key)
+			ws.plain.Set(keys.TxHash.Key(ev.BeginTx.TxHeader.TxHash), bs)
+		}
+
+		bs, err := cdc.MarshalBinaryLengthPrefixed(ev)
+		if err != nil {
+			return err
 		}
+
+		txebs = append(txebs, bs...)
```
silasdavis: what is `txebs`? would be more appropriate as something like `evbs` or `streamEventBytes`

PR #1137, opened 2019-06-13
Do not store empty blocks

seanyoung:


consensus/tendermint/config.go
```diff
@@ -68,7 +68,7 @@ func (btc *BurrowTendermintConfig) Config(rootDir string, timeoutFactor float64)
 
 		// Consensus
 		conf.Consensus.CreateEmptyBlocks = btc.CreateEmptyBlocks
-		conf.Consensus.CreateEmptyBlocksInterval = btc.CreateEmptyBlocksInterval
+		conf.Consensus.CreateEmptyBlocksInterval = btc.CreateEmptyBlocksInterval * time.Second
```
silasdavis: ah good catch!

PR #1138, opened 2019-06-13
Store height offset explicitly, split Tx and Block options

silasdavis:
Instead of using log table - means we always remember a block we have
seen even if we did nothing with it.

Also refactor towards more shared query code via sqlx


silasdavis:
Main file of note here is sqldb.go which github helpfully collapses...

PR #1161, opened 2019-07-02
make test fails in circleci due to out of memory

seanyoung:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -164,7 +164,8 @@ solang: $(SOLANG_GO_FILES)
 
 .PHONY: test
 test: check bin/solc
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -164,7 +164,8 @@ solang: $(SOLANG_GO_FILES)
 
 .PHONY: test
 test: check bin/solc
-	@tests/scripts/bin_wrapper.sh go test ./... ${GOPACKAGES_NOVENDOR}
+# limit parallelism with -p to prevent OOM on circleci
+	@tests/scripts/bin_wrapper.sh go test ./... -p 2
```
silasdavis: Can we have `go test ./... ${GO_TEST_ARGS}`?

PR #1170, opened 2019-07-09
WIP: Preserve origin height

silasdavis:
Currently we will lose original height if repeatedly dumping (more than once).

Making this PR while I'm thinking about it, but we should also test this and we need to think through what guarantees we make about `_height` and `_chain_id` metadata columns in Vent. This is important for consumer who rely on origin chain info.


seanyoung:
Merged in https://github.com/hyperledger/burrow/pull/1176

PR #1171, opened 2019-07-10
Ensure events are dump if no height is provided

seanyoung:


tests/dump/test.sh
```diff
@@ -74,6 +74,12 @@ $burrow_bin dump remote --height $height dump-after-restore.json
 
 kill $burrow_pid
 
+deadcafe=$(grep DEADCAFE dump.json | wc -l)
+if [[ $deadcafe -ne 2 ]]; then
+	echo "DUMP FAILURE -- missing DEADCAFE"
```
silasdavis: I'm pretty sure the first person to experience this error is going to find it somewhat cryptic... Maybe a bit more verbiage about how you are expecting an event with this name

Issue #1183, opened 2019-08-01
Conflict between separate-genesis-doc and pool


ccconnor:
burrow version: 0.27.0
go version: go1.12.4 darwin/amd64

I executed command as below
```
burrow spec -f2 | burrow configure -s- --pool --separate-genesis-doc=genesis.json
```
And then the programme crashed. [Here](https://gist.github.com/ccconnor/fc8ae7b0b68ce40a4e76f886b7d3b93d) is the log.

ccconnor:
I looked into the code, and I found the flag _separate-genesis-doc_  was conflict with the flag _pool_. 

The former assigns _conf.GenesisDoc_ to _nil_ [here](https://github.com/hyperledger/burrow/blob/develop/cmd/burrow/commands/configure.go#L228), while the latter still uses _conf.GenesisDoc_ [here](https://github.com/hyperledger/burrow/blob/develop/cmd/burrow/commands/configure.go#L280).
I tried to comment the assignment expression, and it worded well.

So why does _separate-genesis-doc_ assign _conf.GenesisDoc_ to _nil_? And are there any reliable docs? So far I haven't started Multiple validators either. The quickstart is not so quick.

silasdavis:
@ccconnor you are right, `--pool` and `--separate-genesis-doc` should be mutually exclusive so your invocation should be a usage error. Here is a PR to fix that: https://github.com/hyperledger/burrow/pull/1184/files

>And are there any reliable docs? 

Which docs are currently unreliable? I'm not aware of anywhere that suggests using both of these flags, if that is the only issue. Please do make a PR to fix any faulty docs, or explain the inaccuracy here and we can try to fix them.

A note on `--pool`: we added this flag to help quickly generate local test nets and/or provide a basis for developing configs for bigger networks. It is not meant to be the most flexible tool. For that `burrow configure` and external external scripting (e.g. Kubernetes Helm https://github.com/helm/charts/tree/master/stable/burrow) is likely to be the best way to build networks.

silasdavis:
Okay I've had a little go with this locally and found a bug with `--pool` (not lower casing node IDs - something that has changed in Tendermint) which will be causing you an issue here. See updated #1184.

if you run:

```
burrow spec -f2 | burrow configure -s- --pool
```

that should work fine

ccconnor:
@silasdavis Thanks for you reply. I followed instructions from https://github.com/hyperledger/burrow/blob/develop/docs/quickstart/add-validators.md
[...]

PR #1184, opened 2019-08-02
Fix --pool

silasdavis:
`--pool` will not work with current tendermint because it generates node address with uppercase node IDs (addresses) which, rather annoyingly, is no longer allowed. This fixes that.

Also:
- Allow `--pool` and `--separate-genesis-doc` to work together see #1183 
- It's rather unhelpful to suppress Tendermint logging by default. I have removed this from the default logging config. Excessive Tendermint logs are less of an issue with default empty block settings anyway.

We should add integration tests for `--pool`


compleatang:
FIPF

Issue #1182, opened 2019-08-01
RPC request Names causes panic


ccconnor:
burrow version: 0.27.0
1. start [Multiple validators](https://github.com/hyperledger/burrow/blob/develop/docs/quickstart/multiple-validators.md)
2. curl -s 127.0.0.1:26758/names
This caused panic.

silasdavis:
Thanks!

There's two things here:

1. Although Burrow does not crash from this panic (RPC method panics are recovered from to prevent this) the info server does not recover - this is because an additional panic in the error handling!

2. The /names endpoint on the info service is not correctly wired up. The reason we have not noticed this is that we usually use names over the GRPC service. Not that that is any excuse, the real culprit is that it would appear the info server names endpoints are missing a test (the other endpoints are tested).

silasdavis:
This will be fixed on develop one the PR above is merged...

``` shell
# list all name
curl -s 127.0.0.1:26758/names
# select by regular expression:
curl -s '127.0.0.1:26758/names?regex=Anna'
```

ccconnor:
Nice work!

PR #1188, opened 2019-08-02
Fix name panic

silasdavis:
Adds:

- RPC info integration test for name and names
- Query names by regex
- Fix the panic in rpc lib panic handler!

fixes: #1182


compleatang:
FIPF merge.

PR #1217, opened 2019-08-20
fix: updates example makefiles

kwv:
Signed-off-by: Kris Van de Steeg <kwv@users.noreply.github.com>

### DIFF
--- a/example/basic-app-website/Makefile
+++ b/example/basic-app-website/Makefile
@@ -24,7 +24,7 @@ bin/burrow:
 
 # Generate the chain
 burrow.toml genesis.json:
-	${BURROW_BIN} spec --full-accounts 1 | ${BURROW_BIN} configure --genesis-spec=- --separate-genesis-doc=genesis.json > burrow.toml
+	${BURROW_BIN} spec --full-accounts=1 > genesis-spec.json | ${BURROW_BIN} configure  --separate-genesis-doc=genesis.json --genesis-spec=genesis-spec.json > burrow.toml
 
 # Dump account information to file for app
 account.json: genesis.json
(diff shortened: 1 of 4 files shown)


example/basic-app-website/Makefile
```diff
@@ -24,7 +24,7 @@ bin/burrow:
 
 # Generate the chain
 burrow.toml genesis.json:
-	${BURROW_BIN} spec --full-accounts 1 | ${BURROW_BIN} configure --genesis-spec=- --separate-genesis-doc=genesis.json > burrow.toml
+	${BURROW_BIN} spec --full-accounts=1 > genesis-spec.json | ${BURROW_BIN} configure  --separate-genesis-doc=genesis.json --genesis-spec=genesis-spec.json > burrow.toml
```
seanyoung: Not sure why this is needed. This writes the spec to a file rather than stdout.

Note that the pipe does not do anything any more; in effect, they are two separate commands now.

example/basic-app-website/README.md
```diff
@@ -13,6 +13,7 @@ To run the makefile you will need to have installed:
 - Node.js (the `node` binary)
 - npm (the node package manager)
 - jq (the JSON tool)
+- GO
```
seanyoung: Good point.

example/basic-app-website/Makefile
kwv: I was having issues with `make start_chain` (`burrow.toml` was empty` and went command by command troubleshooting / following the docs.  

this change aligned with the docs:
https://github.com/hyperledger/burrow/blob/develop/docs/tutorials/1-run-full-node.md#configuration

and from the last line of https://github.com/hyperledger/burrow/blob/develop/docs/tutorials/3-deploy-contracts.md

seemed like retaining `genesis-spec.json` was considered helpful.

silasdavis: Thanks for the PR. Those docs show that there is a JS file involved. My feeling here is that is more of an intermediate which represent noise on disk, I'd also like to show how the spec is designed so it can be passed straight through.

Retaining genesis spec is only helpful if you want to make the same chain or, chiefly, if you want to do something more complicated than `burrow spec` allows. I don't think that applies in this case.

PR #1232, opened 2019-08-29
document and improve proposals

seanyoung:


deploy/jobs/jobs_contracts.go
```diff
@@ -438,6 +444,11 @@ func FormulateCallJob(call *def.Call, do *def.DeployArgs, deployScript *def.Play
 	call.Gas = FirstOf(call.Gas, do.DefaultGas)
 
 	// Get address (possibly via key)
+	sourceAddress, err := client.ParseAddress(call.Source, logger)
```
silasdavis: Just check `sourceAddress` is non-empty, if it is we parse it here: https://github.com/hyperledger/burrow/blob/develop/deploy/def/client.go#L682

seanyoung: Good point, this change was entirely unnecessary.

Issue #1244, opened 2019-09-10
Test of vm_test.go don't run all the opcodes when using vm.Call()


madiazp:
To test the call of a smart contract I added : 
```
 func TestBigModExp(t *testing.T) {
	cache := NewState(newAppState(), blockHashGetter)
	ourVm := NewVM(newParams(), crypto.ZeroAddress, nil, logger)
	account1 := newAccount(cache, "1")
	account2 := newAccount(cache, "101")

	var gas uint64 = 10000
	fmt.Printf("Testing bigModExp, inputs \n")
	bc := hex.MustDecodeString("608060405234801561001057600080fd5b50610109806100206000396000f3fe6080604052348015600f57600080fd5b506004361060285760003560e01c8063783ce45814602d575b600080fd5b606a60048036036060811015604157600080fd5b810190808035906020019092919080359060200190929190803590602001909291905050506080565b6040518082815260200191505060405180910390f35b60008060405160208152602080820152602060408201528560608201528460808201528360a082015260208160c083600060056107d05a03f1915080519250508060c957600080fd5b81915050939250505056fea265627a7a723158207adae9fda357a576526c1acb8c3c6b3866b1b7831b82d833cd31c8190d23b5e964736f6c634300050b0032")
	input := hex.MustDecodeString("783ce458000000000000000000000000000000000000000000000000000000000000000b00000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000009")
	//addr := crypto.Address{11, 101}
	ourVm.debugOpcodes = true
	out1, err := ourVm.Call(cache, NewNoopEventSink(), account1, account2, bc, input, 0, &gas)

	fmt.Printf("output: %x", out1)

	require.NoError(t, err)

}
```

in the [vm_test.go](https://github.com/hyperledger/burrow/blob/develop/execution/evm/vm_test.go) file. The output was:

`Testing bigModExp, inputs 
(1) (C47907ABD2A80492CA9388B05C0E382518FF3960) FE9362E30831F55EEB1F0D837CD5F98848386EF9 (code=297) gas: 10000 (d) 783CE458000000000000000000000000000000000000000000000000000000000000000B00000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000009
[...]

PR #1246, opened 2019-09-10
Adding the 5th Precompiled Contract to Natives.

madiazp:
Signed-off-by: Matias Díaz matias.dpz@gmail.com
The implementation is mostly code from the go-ethereum project. Maybe It's a problem for the license but the implementation of the contract has to be exactly as in the ethereum VM to work with sol contracts. And it's seems that there is a lot of go-ethereum code in burrow anyway.

The LeftPad and RighPad function are from the burrow/binary package so maybe there is no problem.

### DIFF
--- a/execution/native/precompiles.go
+++ b/execution/native/precompiles.go
@@ -2,6 +2,7 @@ package native
 
 import (
 	"crypto/sha256"
+	"math/big"
 
 	"github.com/hyperledger/burrow/binary"
 	"github.com/hyperledger/burrow/crypto"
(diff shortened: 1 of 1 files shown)


madiazp:
[I couldn't test it with test files](https://github.com/hyperledger/burrow/issues/1244). I tested it in my fabric-cc-evm environment and it worked fine. If the issue above is solved I have no problem to adding a test.

execution/evm/native.go
```diff
@@ -140,3 +151,139 @@ func identityFunc(state Interface, caller crypto.Address, input []byte, gas *uin
...
+	if mod.BitLen() == 0 {
+		// Modulo 0 is undefined, return zero
+		return LeftPadBytes([]byte{}, int(modLen)), nil
+	}
+	return LeftPadBytes(base.Exp(base, exp, mod).Bytes(), int(modLen)), nil
+
+}
+
+// RequiredGas returns the gas required to execute the pre-compiled contract.
+func bigModExpGasRequire(input []byte) uint64 {
```
silasdavis: For our purposes I think we can just elide this complex gas calculation. Gas is something we need to revisit at some stage, but in Burrow right now all gas does is bound computation. A unitary gas cost for each operation is sufficient for this.

silasdavis:
Fabulous! And you integrated with the new natives. Many Thanks.

PR #1289, opened 2019-10-09
Switch to life fork that depends on versioned go-wagon

silasdavis:


silasdavis:
:smiley_cat:

PR #1293, opened 2019-10-10
Split out metadata cache and interfaces

silasdavis:
Hooray another cache I hear you say.

We let this slide originally because it was convenient. But it forces downstream to implement no-op metadata interfaces even though they are not needed by the EVM. It's cleaner for us to keep these separate and only have the EVM ask for what it needs.


silasdavis:
:smiley_cat:

PR #1338, opened 2020-02-14
Replace wasm interface with ewasm

seanyoung:


docker-compose.yml
```diff
@@ -1,9 +1,13 @@
 version: "3"
 services:
   db:
-    image: postgres:11-alpine
+    image: postgres:9
     ports:
-      - 5432
+      - "5432:5432"
+    environment:
+      - POSTGRES_HOST_AUTH_METHOD=trust
```
silasdavis: sorry we only want the two lines above

gregdhill (approved):

PR #1360, opened 2020-03-19
Fixed a bug for operator SDIV.

dshe-certik:
According to [Solidity Docs](https://solidity.readthedocs.io/en/latest/types.html): "In Solidity, division rounds towards zero. This mean that `int256(-5) / int256(2) == int256(-2)`."
In `big.Int`, `Div` implements [Euclidean division](https://en.wikipedia.org/wiki/Euclidean_division) in which the direction of rounding depends on the sign of denominator.

| Example | Solidity/EVM | big.Int.Div() |
| :-: | :-: | :-: |
| 7/3 | 2 | 2 |
| 7/(-3) | -2 | -2 |
| (-7)/3 | -2 | -3 |
| (-7)/(-3) | 2 | 3 |

`case SDIV` in contract.go needs to be modified if we want to keep the same behavior as Solidity/EVM's.

### DIFF
--- a/execution/evm/contract.go
+++ b/execution/evm/contract.go
@@ -106,7 +106,7 @@ func (c *Contract) execute(st engine.State, params engine.CallParams) ([]byte, e
 				stack.Push(Zero256)
 				c.debugf(" %v / %v = %v\n", x, y, 0)
 			} else {
-				div := new(big.Int).Div(x, y)
+				div := new(big.Int).Quo(x, y)
 				res := stack.PushBigInt(div)
 				c.debugf(" %v / %v = %v (%v)\n", x, y, div, res)
 			}


silasdavis:
@dshe-certik thanks so much for this. I'd like to also add a unit test for this at the same time as fixing. If you are willing/able to do this then great and I can give you some pointers (it should be in `evm_test.go`), otherwise I am happy to do it.

One thing we do need in order to accept contributions is DCO signoff. Please just follow instructions here: https://github.com/hyperledger/burrow/pull/1360/checks?check_run_id=519925126 and I will make sure your original commit is merged as part of this change.

dshe-certik:
> @dshe-certik thanks so much for this. I'd like to also add a unit test for this at the same time as fixing. If you are willing/able to do this then great and I can give you some pointers (it should be in `evm_test.go`), otherwise I am happy to do it.
> 
> One thing we do need in order to accept contributions is DCO signoff. Please just follow instructions here: https://github.com/hyperledger/burrow/pull/1360/checks?check_run_id=519925126 and I will make sure your original commit is merged as part of this change.

Thanks for the suggestion! I will do it.

Actually I just found a similar problem with `SMOD` and another problem with `JUMP` and `JUMPI`. I'll fix them and added unit tests.

silasdavis:
> Actually I just found a similar problem with `SMOD` and another problem with `JUMP` and `JUMPI`. I'll fix them and added unit tests.

Fabulous! :pray:

PR #1365, opened 2020-04-05
Build a burrow debug binary and include it in the docker image with the

silasdavis:
name 'burrow-debug'


silasdavis:
:smiley_cat:

PR #1367, opened 2020-04-06
Fixed bugs related to operators in EVM

dshe-certik:
#### `SDIV`
According to [Solidity Docs](https://solidity.readthedocs.io/en/latest/types.html): "In Solidity, division rounds towards zero. This mean that `int256(-5) / int256(2) == int256(-2)`."
In `big.Int`, `Div` implements [Euclidean division](https://en.wikipedia.org/wiki/Euclidean_division) in which the direction of rounding depends on the sign of denominator.

| Example | EVM/big.Int.Quo() | big.Int.Div() |
| :-: | :-: | :-: |
| 7/3 | 2 | 2 |
| 7/(-3) | -2 | -2 |
| (-7)/3 | -2 | -3 |
| (-7)/(-3) | 2 | 3 |

#### `SMOD`
According to [Ethereum Yellow Paper](https://ethereum.github.io/yellowpaper/paper.pdf), we should use `big.Int.Rem()` rather than `big.Int.Mod()` for operator `SMOD`.
| Example | EVM/big.Int.Rem() | big.Int.Mod() |
| :-: | :-: | :-: |
| 7 mod 3 | 1 | 1 |
| 7 mod (-3) | 1 | 1 |
| (-7) mod 3 | -1 | 2 |
| (-7) mod (-3) | -1 | 2 |

#### `SIGNEXTEND`
Fixed incorrect sign bit and prevented interger overflow in [this test case](https://github.com/ethereum/tests/blob/develop/VMTests/vmArithmeticTest/signextend_BigByteBigByte.json).

#### `MSOTRE8`
Fixed interger overflow problem in [this test case](https://github.com/ethereum/tests/blob/develop/VMTests/vmIOandFlowOperations/mstore8WordToBigError.json).

#### `JUMPDEST`
Fixed the bug that `JUMP` and `JUMPI` can jump to destinations inside `PUSH` data.
[Ethereum wiki](https://github.com/ethereum/wiki/wiki/Subtleties):
"`JUMP` and `JUMPI` instructions are only allowed to jump onto destinations that are (1) occupied by a `JUMPDEST` opcode, and (2) are **not inside `PUSH` data**."


seanyoung:
Thanks for fixes, they really improve things. It would be nice to see the divide and module changes tested in Solidity so it is clearer it actually works as intended.

I've added a similar test to Solang: https://github.com/hyperledger-labs/solang/commit/c3b56d064db6e41d9a821639c8c11d2ae1fedae8

The solidity from that commit should go into a solidity deploy test here: https://github.com/hyperledger/burrow/tree/master/tests/jobs_fixtures

dshe-certik:
> Thanks for fixes, they really improve things. It would be nice to see the divide and module changes tested in Solidity so it is clearer it actually works as intended.
> 
> I've added a similar test to Solang: [hyperledger-labs/solang@c3b56d0](https://github.com/hyperledger-labs/solang/commit/c3b56d064db6e41d9a821639c8c11d2ae1fedae8)
> 
> The solidity from that commit should go into a solidity deploy test here: https://github.com/hyperledger/burrow/tree/master/tests/jobs_fixtures

I have added tests in Solidity for `SDIV` and `SMOD`. Please let me know if I should do it in another way or if there's any other problems. Thanks!

execution/evm/contract.go
```diff
@@ -862,6 +862,22 @@ func (c *Contract) jump(code []byte, to uint64, pc *uint64) error {
 	return nil
 }
 
+// isInsidePushData checks if the operator code(n) is inside push data
```
silasdavis, line 865:
```suggestion
// isInsidePushData checks if the symbol at code(n) is inside push data i.e. is data for the stack and not an opcode
```

execution/evm/contract.go
```diff
@@ -862,6 +862,22 @@ func (c *Contract) jump(code []byte, to uint64, pc *uint64) error {
...
+func isInsidePushData(code []byte, n uint64) bool {
+	if uint64(len(code)) <= n {
+		return false
+	}
+	i := uint64(0)
+	for i < n {
```
silasdavis, line 871: This could get expensive for long code. A simple improvement on the average case would be to scan from `n` backwards to the first `PUSH<N>` opcode and see whether n falls within [i, i+N]. The worst case (no pushes) is the same, but there are usually plenty of pushes.

I note go-ethereum amortises this work by creating a bitmap that partitions each symbol in the code into code/not code (https://github.com/ethereum/go-ethereum/blob/master/core/vm/analysis.go). I think I prefer the simple scanning approach. I suspect there are enough pushes in real solidity code that the 'count back from n' algorithm is good enough in practice.

Another ahead-of-time approach would be to store an array of offsets for all `PUSH`es in the code, then perform a binary search to find the nearest PUSH to the left.

What do people think about this? Should we bite the bullet and eat the cost for this at contract deploy time. If so which of the two approaches above seems best? If we are doing a linear pass of the code then we also have an opportunity to capture other metadata, not sure if there is anything that might be useful there.

silasdavis (commented):
@dshe-certik we are very grateful for this contribution. It fixes some significant issues and I really appreciate you taking the time to put a bow on it with the tests you have written.

The only substantive issue I see is with the new check that a `JUMPDEST` is really a jump destination and not just data (much needed!) since it could get quite expensive. Go-ethereum bothers to do something about this, and I am inclined to think we should too.

The approach I would favour would be adding a sorted array of `PUSH` locations on to our `Account` model and initialising that array here: https://github.com/hyperledger/burrow/blob/master/execution/native/account.go#L40-L83.

@seanyoung is there any way we could make this generic/useful for WASM code too? I was wondering if we had something like:

```
type IndexCode byte

const (
  Pushes IndexCode = iota
  ...
)

type Account {
  ...
  CodeIndices map[IndexCode][]uint64 // we might need to be careful with maps (I _think_ our protobuf options should make their serialisation deterministic, otherwise we can just use a slice of structs)
}
```

Then we could have `CodeIndices[Pushes] = []uint64{10, 34, 50, ...}`. If we decide we would like other types of index (perhaps a WASM specific one) then we could add them here rather than making another code-type specific field.

This would give us `O(log(n))` lookups for less space than the bitmap approach (provided `Exp[number of pushes] < len(code)/8`).

---

Ultimately I am happy to merge this with the slightly improved naive implementation and we can fix it up in a subsequent PR (though I think we should _probably_ do this before the next release) but let's have a brief discussion now.

@dshe-certik I'd love to have you as a more regularly contributor and if you are interested in maintainership then we are on the look out for people. Would be great to see you in https://chat.hyperledger.org/channel/burrow-contributors or ping me at silas at monax dot io regarding maintainership.

dshe-certik:
> @dshe-certik we are very grateful for this contribution. It fixes some significant issues and I really appreciate you taking the time to put a bow on it with the tests you have written.
> 
> The only substantive issue I see is with the new check that a `JUMPDEST` is really a jump destination and not just data (much needed!) since it could get quite expensive. Go-ethereum bothers to do something about this, and I am inclined to think we should too.
> 
> The approach I would favour would be adding a sorted array of `PUSH` locations on to our `Account` model and initialising that array here: https://github.com/hyperledger/burrow/blob/master/execution/native/account.go#L40-L83.
> 
> @seanyoung is there any way we could make this generic/useful for WASM code too? I was wondering if we had something like:
> 
> ```
> type IndexCode byte
> 
> const (
>   Pushes IndexCode = iota
>   ...
> )
> 
> type Account {
>   ...
>   CodeIndices map[IndexCode][]uint64 // we might need to be careful with maps (I _think_ our protobuf options should make their serialisation deterministic, otherwise we can just use a slice of structs)
> }
> ```
> 
> Then we could have `CodeIndices[Pushes] = []uint64{10, 34, 50, ...}`. If we decide we would like other types of index (perhaps a WASM specific one) then we could add them here rather than making another code-type specific field.
> 
> This would give us `O(log(n))` lookups for less space than the bitmap approach (provided `Exp[number of pushes] < len(code)/8`).
> 
> Ultimately I am happy to merge this with the slightly improved naive implementation and we can fix it up in a subsequent PR (though I think we should _probably_ do this before the next release) but let's have a brief discussion now.
> 
> @dshe-certik I'd love to have you as a more regularly contributor and if you are interested in maintainership then we are on the look out for people. Would be great to see you in https://chat.hyperledger.org/channel/burrow-contributors or ping me at silas at monax dot io regarding maintainership.

@silasdavis Thanks for the invitation! I would be very happy to learn more about Burrow. I tried the link (https://chat.hyperledger.org/channel/burrow-contributors) but it kept loading and never finished...

execution/evm/contract.go
```diff
@@ -862,6 +862,22 @@ func (c *Contract) jump(code []byte, to uint64, pc *uint64) error {
...
+		if op := OpCode(code[i]); op >= PUSH1 && op <= PUSH32 {
+			i += uint64(op - PUSH1 + 2)
```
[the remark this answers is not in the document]
dshe-certik: Yeah it makes the code clearer. Fixed.

execution/evm/contract.go
```diff
@@ -164,9 +164,9 @@ func (c *Contract) execute(st engine.State, params engine.CallParams) ([]byte, e
 			c.debugf(" %v ** %v = %v (%v)\n", x, y, pow, res)
 
 		case SIGNEXTEND: // 0x0B
-			back := stack.Pop64()
+			back := stack.PopBigInt().Uint64()
 			if back < Word256Bytes-1 {
-				bits := uint(back*8 + 7)
+				bits := uint(back*8 + 8)
```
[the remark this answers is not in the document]
dshe-certik: I rewrote this line as `bits := uint((back + 1) * 8)`. I didn't understand your explanations on changing `SignExtend()`, so I didn't make further modifications.

execution/evm/contract.go
dshe-certik: Updated.

execution/evm/contract.go
dshe-certik: I agree this is expensive, but I don't think the 'count back from n' algorithm would work.
For example, in `PUSH1, PUSH1, JUMPDEST` the second `PUSH1` is data rather than opcode. If we simply scan backwards `JUMPDEST` would be treated as data as the second `PUSH1`.
I think the other algorithm you mentioned that stores an array of offsets for all `PUSH`es is a good one.

silasdavis:
@dshe-certik 

> I tried the link (https://chat.hyperledger.org/channel/burrow-contributors) but it kept loading and never finished...

rocket chat can be a little flaky... you will need to create a linux foundation account an login. Try browsing to chat.hyperledger.org/, clicking general, and it should show you a button prompting login

execution/evm/contract.go
silasdavis: Yes you are right!

However we can:

Define `P(n)` to be the index of the closest previous push that captures `n`, that is:

`P(n)` defined as the largest integer such that `P(n) < n` and `code[P(n)] = PUSH<N>` where `n-P(n) <= N` or else null.

Then for a `JUMPDEST` at `n` we can search for`P(n)` then ask 'is this a real push?' by searching for `P(P(n))`. If `P(P(n))` is null then we are done and `P(n)` must be real, and therefore `JUMPDEST` is just data.

However if `P(P(n))` exists then we must ask i's `P(P(n))` a real push'. We can carry on searching backwards recursively like this. It will terminate when we there is no previous capturing push. In a chain like this the leftmost push is the first 'real one', the odd-numbered ones are just data so eventually you need to check if the rightmost real push captures n (since P(n) may be data, but P(P(n)) might still capture n).

So having thought that through it seems like quite a bit of work for a still sub-optimal solution!

dshe-certik:
> @dshe-certik
> 
> > I tried the link (https://chat.hyperledger.org/channel/burrow-contributors) but it kept loading and never finished...
> 
> rocket chat can be a little flaky... you will need to create a linux foundation account an login. Try browsing to chat.hyperledger.org/, clicking general, and it should show you a button prompting login

Thanks! I joined the channel!

execution/evm/contract.go
dshe-certik: Cool! I'll implement this.

silasdavis:
I'd really like to merge this but I don't think we can really live with a linear scan.

Also on reflection I think that pushes are so frequent in EVM code that the space saving will be minimal or negative for the array of push locations + binary search algorithm i proposed and the cost of doing the binary search every time we jump prohibitive.

I think go-ethereum has gone the right way on this. We should just add an additonal 1/8th of the code size by providing a bitmap lookup for 'is this code?'. However we should not (and can not) copy GPL code from go-ethereum.

I suggest we use this library: https://github.com/tmthrgd/go-bitset which both gives a dead simple implementation of an arbitrary `len(code)` bitset but also does some low level optimisation to make it even faster than a straight implementation in Go, with the added benefit it is clearly a different implementation to geth: https://github.com/tmthrgd/go-bitset.

Is this something you are willing/able to work on? I can take a look but I am not sure when. One thing I can do ASAP if you do want to work on it is update our protobuf so the account object has (this is fiddly an unedifying):

```go
CodeSegment Bitset
```

Then it would just be a case of implementing the filling of this bitmap in `InitCode`, integrating the function to check the bitmap, and adding a test.

execution/evm/contract.go
silasdavis: Oops, see my comment below - I do not think this is worth doing - I think we should do the bitmap.

silasdavis:
Okay so I have implemented the bitset here: https://github.com/hyperledger/burrow/pull/1369, if you rebase on master once that is in there you can use that to check, so your `JUMPDEST` check will become:

```go
opcodeBitset := mustGetAccount(st.CallFrame, maybe, params.Callee).OpcodeBitset
if dest != JUMPDEST || !opcodeBitset.IsSet(to) {
...
}
```

dshe-certik:
> Okay so I have implemented the bitset here: #1369, if you rebase on master once that is in there you can use that to check, so your `JUMPDEST` check will become:
> 
> ```go
> opcodeBitset := mustGetAccount(st.CallFrame, maybe, params.Callee).OpcodeBitset
> if dest != JUMPDEST || !opcodeBitset.IsSet(to) {
> ...
> }
> ```

Now that we have `OpcodeBitset`, we should always call `initEVMCode()` before calling `vm.Execute()`. Is this correct?

silasdavis:
> Now that we have OpcodeBitset, we should always call initEVMCode() before calling vm.Execute(). 

> Is this correct? 
no

>Is this guaranteed?

Yes, but you don't need to make any additional call to `initEVMCode()` - it is already called by the contract creation machinery.

Also I just pushed an update: https://github.com/hyperledger/burrow/pull/1373 to make this a bit less leaky. The code you want for the `JUMPDEST` check is now:

```go
account := mustGetAccount(st.CallFrame, maybe, params.Callee)
if dest != JUMPDEST || !account.IsOpcodeAt(to) {
...
}
```

dshe-certik:
> ```go
> IsOpcodeAt
> ```

This is somewhat different from my understandings. I thought when we deploy a contract, we will call `vm.Execute()` to execute the bytecode, which could have `JUMPDEST` in it, and then call `initEVMCode()` to update the callee account with returned code.
There's a test named `MsgSender` in evm_test.go. It does the things I described above.
Did I miss anything?

silasdavis:
Your understanding is correct. I hadn't considered the case when we are running the contract creation code. In that case the `OpcodeBitset` is not initialised.

I wasn't entirely happy with putting that field on `Account`. Since we run code in contexts without an account and it is EVM specific. Better to have the type system stop you from having unintialised code.

I've attempted to do that here: https://github.com/hyperledger/burrow/pull/1374/files#diff-a657cd90f7f9b89ca0d3edf9bc6086beR26-R31

By introducing a wrapper type around the bytecode that includes the bitset we expect to exist. That way whenever you have a non-nil `EVMCode` you can expect it to have `OpcodeBitset`: https://github.com/hyperledger/burrow/pull/1374/files#diff-7f728301f937f2d95a459df952bd17bcR854-R856

This works but it does break the upgradeability of our chains via dump/restore. All of these changes are potentially state hash breaking so we have to live with that anyway.

I am on the fence whether we should store this bitset in state or whether we should just generate once (via a single pass over the code) when we actually call `Execute`. The latter means we don't have to change our account state at all...

silasdavis:
Okay so I much prefer the simpler version: https://github.com/hyperledger/burrow/pull/1375

If it still seems like I good idea in the morning I'll try and get that merged and it should give you what you need and we can get this in.

silasdavis:
@dshe-certik sorry for the wall of text for you to come back to. We have know merged a working solution for the jump dest issue to master. If you rebase off master and remove your original helper function your other changes should be good to go.

Let me know if you have any comments or queries on the changes there (see https://github.com/hyperledger/burrow/pull/1375)

dshe-certik:
> @dshe-certik sorry for the wall of text for you to come back to. We have know merged a working solution for the jump dest issue to master. If you rebase off master and remove your original helper function your other changes should be good to go.
> 
> Let me know if you have any comments or queries on the changes there (see #1375)

I think it's a pretty good solution. I have modified the code based on your updates.

silasdavis:
Hooray, thanks!

PR #1369, opened 2020-04-13
Opcode bitmap

silasdavis:
This is just some prep for @dshe-certik to add efficient jump dest validation. Mostly code regen. Source code changes in first commit: https://github.com/hyperledger/burrow/pull/1369/commits/9630cf7038fbbb93bc0fbda2a086b9fc66da201a


seanyoung (approved):
I like the idea of the bitset for valid opcodes, looks good.

Another way of doing this is to have a sorted list of valid jump destinations. That might be smaller than the bitset of opcodes. It would require binary search rather than testing bitset element, though.

silasdavis:
> I like the idea of the bitset for valid opcodes, looks good.
> 
> Another way of doing this is to have a sorted list of valid jump destinations. That might be smaller than the bitset of opcodes. It would require binary search rather than testing bitset element, though.

yep, I discussed this extensively (mostly with myself) on #1366, but I think ultimately there are so may `PUSH`es that it may not even save space and probably isn't even worth the repeated binary search for each jump...

PR #1416, opened 2021-01-19
[fix]execution/evm/abi: check data length before decoding Address and Bytes with ABI

Tri-stone:
Check data length before decoding `Address` and `Bytes` with ABI, or it will panic by decoding an `Address` or `Bytes` value.


silasdavis:
Thanks!

PR #1421, opened 2021-02-01
WIP: WASM cross dispatch

silasdavis:
Things to discuss:

- How to test calls/use solang (fancy handling this @seanyoung ?)
- Calling convention translation? Does eWASM handle this in compiled code or do we need to do translation? Would we ever make use of the Burrow-hosted ABIs?
- Error handling: is panic/recover the way to go here or should we have something more explicit? Should we do a bit more bounds and sanity checking on bad input code (ultimately we have panic/recover over the executors anyway but still)?
- What about other call types; `STATICCALL` `CALLCODE` `DELEGATECALL` - I have abstracted the logic that handles these in the EVM, not sure if solang is able to create code that makes use of them?
- Future PR probably but we need event support.


seanyoung:
> How to test calls/use solang (fancy handling this @seanyoung ?)

I'm happy to write some test cases. 

>    Calling convention translation? Does eWASM handle this in compiled code or do we need to do translation? Would we ever make use of the Burrow-hosted ABIs?

Not needed. The arguments are encoded/decoded in wasm.

>    Error handling: is panic/recover the way to go here or should we have something more explicit? Should we do a bit more bounds and sanity checking on bad input code (ultimately we have panic/recover over the executors anyway but still)?

I'm not sure when we would encounter this. From burrow's perspective, the call specifies the calldata bytes, and this needs to be passed verbatim to the called contract.

>   What about other call types; STATICCALL CALLCODE DELEGATECALL - I have abstracted the logic that handles these in the EVM, not sure if solang is able to create code that makes use of them?

Right now solang will generate these if you call the `staticcall()` or `delegatecall()` on an address type.

>    Future PR probably but we need event support.

This is separate from call support.

PR #1422, opened 2021-02-02
Implement cross-engine call dispatch for eWASM

silasdavis:
- Implement the eWASM `call` syscall to support calling contracts

- Wire up WASM VM as a Disptacher so it can call EVM and Natives
(including Precompiles) as well as other WASM contracts.

- To support 128-bit wide (and 256-bit wide EVM!) values use big.Int for
Value and Gas. Note the Account model currently still uses uint64 so
outside of the execution machinery we still throw an error if a balance
would overflow uint64. This paves the way for relaxing this requirement
and making account balances etc.

- exec.CallData `Value` and `Gas` are now the big-endian big.Int bytes
rather than uint64 to support the above.

- EVM and WASM now share calling and some other code. EVM now more
strictly throws errors that are not ExecutionReverted becuase it
probably always should have and makes it more consistent with WASM.

WASM uses panics to transmit its errors. If we have errors that should
not halt execution (AKA 'non-trapping') we should probably explicitly encode
this fact into the coded errors and act accordingly and consistenly in
WASM and EVM


execution/wasm/contract.go
```diff
@@ -84,6 +85,12 @@ func (ctx *context) ResolveFunc(module, field string) lifeExec.FunctionImport {
 
 	switch field {
 	case "call":
+		fallthrough
+	case "callCode":
+		fallthrough
+	case "callDelegate":
+		fallthrough
+	case "callStatic":
```
silasdavis, line 93: you can comma-delimiter these values under a single case statement, think that is cleaner/more/less likely to accidentally split cases.

silasdavis (commented):
Nit

execution/wasm/contract.go
seanyoung: right, that is nicer. Something for the next ewasm pr

PR #1427, opened 2021-02-04
Wasm missing interface calls

yoongbok-lee:
Implemented:
 - `create`
 - `selfdestruct`

Placeholders:
 - `getTxGasPrice`
 - `getBlockDifficulty`

Fixed:
 - `call` opcodes value read


execution/wasm/contract.go
```diff
@@ -220,13 +222,126 @@ func (ctx *context) ResolveFunc(module, field string) lifeExec.FunctionImport {
...
+
+			err = ctx.state.CallFrame.UpdateAccount(acc)
+			if err != nil {
+				panic(err)
+			}
+			err = ctx.state.CallFrame.RemoveAccount(ctx.params.Callee)
+			if err != nil {
+				panic(err)
+			}
+			return 0
```
seanyoung, line 330: Execution should halt here. Use panic(errors.Codes.None)

execution/wasm/contract.go
```diff
@@ -220,13 +222,126 @@ func (ctx *context) ResolveFunc(module, field string) lifeExec.FunctionImport {
...
+		}
+
+	case "getTxGasPrice":
+		return func(vm *lifeExec.VirtualMachine) int64 {
+			resultPtr := int(vm.GetCurrentFrame().Locals[0])
+
+			// set it to 1
+			binary.BigEndian.PutUint64(vm.Memory[resultPtr:resultPtr+16], 1)
+			return 0
```
seanyoung, line 292: The "price" is a 128 bit value, no 64 bit.

seanyoung, line 292: return Success; (also the other functions)

execution/wasm/contract.go
```diff
@@ -220,13 +222,126 @@ func (ctx *context) ResolveFunc(module, field string) lifeExec.FunctionImport {
...
+
+			return Success
+		}
+
+	case "getBlockDifficulty":
+		return func(vm *lifeExec.VirtualMachine) int64 {
+			resultPtr := int(vm.GetCurrentFrame().Locals[0])
+
+			// set it to 1
+			binary.BigEndian.PutUint64(vm.Memory[resultPtr:resultPtr+32], 1)
```
seanyoung, line 282: block difficulty is a 256 bit value

execution/wasm/contract.go
```diff
@@ -72,6 +73,7 @@ type context struct {
 	code       []byte
 	output     []byte
 	returnData []byte
+	sequence   uint64
```
seanyoung, line 76: Is the sequence populated from the account on execution?

execution/wasm/contract.go
seanyoung: You're just writing a 64 bit value here

execution/wasm/contract.go
```diff
@@ -220,13 +222,126 @@ func (ctx *context) ResolveFunc(module, field string) lifeExec.FunctionImport {
...
+			resultPtr := uint32(vm.GetCurrentFrame().Locals[3])
+
+			// TODO: is this guaranteed to be okay? Should be avoid panic here if out of bounds?
+			value := bin.BigIntFromLittleEndianBytes(vm.Memory[valuePtr : valuePtr+ValueByteSize])
+
+			var data []byte
+			copy(data, vm.Memory[dataPtr:dataPtr+dataLen])
+
+			ctx.sequence++
+			nonce := make([]byte, txs.HashLength+8)
```
seanyoung, line 239: The sequence of the account should be updated too

seanyoung (changes requested):
Please squash your changes to a single commit and include a Signed-off-by line

This is what is causing the DCO check to fail on CI.

Thanks for you work.

execution/wasm/contract.go
yoongbok-lee: Right, I changed it to `copy(vm.Memory[resultPtr:resultPtr+32], bin.One256[:])`

execution/wasm/contract.go
yoongbok-lee: Fixed!

yoongbok-lee: How should I get a 128 bit 1? I've changed to `copy(vm.Memory[resultPtr:resultPtr+16], bin.LeftPadBytes([]byte{1}, 16))`

execution/wasm/contract.go
yoongbok-lee: Fixed!

execution/wasm/contract.go
yoongbok-lee: It was taken from execution sequence from EVM, 
```
func (vm *EVM) SetNonce(nonce []byte) {
	vm.options.Nonce = nonce
	vm.sequence = 0
}
```
since Nonce is also provided through initial wasm call, we can include account sequence there.

execution/wasm/contract.go
yoongbok-lee: I think if my understanding is correct, I don't think we need to update the account sequence (since it should be handled outside the wasm call) after execution sequence number modification

execution/wasm/contract.go
seanyoung: wasm is little endian. This should be bin.RightPadBytes()

yoongbok-lee: Got it, fixed

seanyoung (approved):

yoongbok-lee:
Anything else I should do? should I add in some tests?

seanyoung:
Waiting for @silasdavis to review

seanyoung:
Some tests would be good though!

yoongbok-lee:
Found some bugs while writing tests. I'm almost done fixing them, so I'll push again once I'm done

silasdavis:
Please bear with me, this is on my stack

seanyoung:
I've got some changes which are waiting for this to be merged.

I asked @silasdavis a week ago to review this. Let's merge now and if there are any issues we can open a new PR.

silasdavis:
LGTM. I think it could use some more tests, but it's still quite young code.

PR #1429, opened 2021-02-05
Docker image should start testnet by default

seanyoung:


Dockerfile
```diff
@@ -43,4 +44,4 @@ EXPOSE 26658
 EXPOSE 10997
 
 USER $USER:$USER
-ENTRYPOINT [ "burrow" ]
+ENTRYPOINT [ "burrow",  "start" ]
```
compleatang, line 47: This is a non-standard use of entrypoint, AFAIK. 

Better to either change this line to a `CMD` or to add the `start` as a `CMD` and keep the entrypoint as is.

compleatang (changes requested):
This change will break a lot of builds // charts, can we get a bit more clarity about why we would want to make this change?

seanyoung:
The idea is that if you run ``docker run hyperledger/burrow:latest`` you get a development chain up and running for free. Nice and easy for newcomers.

Having said that I appreciate that changing entrypoint might break stuff. 

So what's the idea of having burrow as the entrypoint? If run ``docker run hyperledger/burrow:latest``, it shows me the --help output.

compleatang:
I get making it easier for new users, but we're gonna break all the production runtime configs that use this docker container with this change as is because properly configured runtime configs will have `start` or `vent` or a tool as the command. I would suggest that we just tweak what you have to:

```
ENTRYPOINT ["burrow"]
CMD ["start"]
```

That would get the same effect you mention but still allow things like `docker run hyperledger/burrow vent` and won't break folks' charts or other runtime artifacts.

seanyoung:
Great idea, thanks @compleatang

It works fine.

compleatang (approved):

compleatang:
Great! LGTM.

PR #1445, opened 2021-03-10
Ensure `docker run hyperledger/burrow` works

seanyoung:
burrow.toml is missing


compleatang:
LGTM

PR #1450, opened 2021-03-11
Make js package work with numbers > 53 bit

seanyoung:

### DIFF
--- a/js/src/lib/utils/convert.ts
+++ b/js/src/lib/utils/convert.ts
@@ -1,6 +1,6 @@
 import BN from 'bn.js';
 
-export const recApply = function<A, B> (func: (input: A) => B, args: A | A[]): B | B[] {
+export const recApply = function <A, B>(func: (input: A) => B, args: A | A[]): B | B[] {
   if (Array.isArray(args)) {
     let next: any = [];
     for (let i = 0; i < args.length; i++) {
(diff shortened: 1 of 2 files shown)


js/src/lib/utils/convert.ts
```diff
@@ -34,7 +34,16 @@ export const bytesTA = function (arg: string) {
 }
 
 export const numberTB = function (arg: BN) {
-  return arg.toNumber();
+  let res: BN | number;
+  try {
+    // number is limited to 53 bits, BN will throw Error
+    res = arg.toNumber();
+  }
+  catch {
```
silasdavis, line 42: Any chance we could be more specific about the catch here? If it would rely on something fragile like matching a string then don't bother.

silasdavis (approved):
This is stupid, there is no reasonable application that could require more than 53 bits.

js/src/lib/utils/convert.ts
seanyoung: Yes, I had considered that but bn.js simply does a `throw Error(''Number can only safely store up to 53 bits')` 
see https://github.com/indutny/bn.js/blob/master/lib/bn.js#L547  and https://github.com/indutny/bn.js/blob/master/lib/bn.js#L5

PR #1451, opened 2021-03-11
Fix vent EVM-to-SQL integral number type mapping

silasdavis:
Also add dev builds for JS


seanyoung:
Looks good but tests don't pass

PR #1452, opened 2021-03-12
Add continuity test to vent

silasdavis:
BlockConsumerConfig parameterises retries and backoff.

Continuity test ensures vent consumers read an exact cover of block
batches (non-overlapping and complete)

Also:

- Made logging less painful to use for debug etc
- Support block info from Ethereum


compleatang:
LGTM

Issue #1453, opened 2021-03-15
OS portability could be improved


bzp99:
**Describe the Bug**
Not all OSs keep `bash` under `/bin/`, for example OpenBSD puts binaries installed from ports to `/usr/local/bin/` (which is very useful). Also, `head`’s `-c` flag is not portable.

**To Reproduce**
_(not applicable)_

**Expected Behavior**
Burrow is easy to build on most OSs.

**Technical Details**
_(not applicable)_

**Additional Context**
I will soon create a pull request to address this. I create this issue following  the [contribution guide](https://github.com/hyperledger/burrow/blob/master/.github/CONTRIBUTING.md).

bzp99:
Resolved by #1454

PR #1454, opened 2021-03-15
Improve OS portability

bzp99:
Tiny fixes for the issues mentioned in #1453, namely:

* use `/usr/bin/env bash` instead of calling `/bin/bash` directly (which may not always be there)
* use `dd` instead of `head -c`, which is not portable
* use `head -n` instead of the historical `head -<count>` syntax
* do not use non-portable `-T` flag with `install`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -169,7 +169,7 @@ build_burrow_debug:
 .PHONY: install
 install: build_burrow
 	mkdir -p ${BIN_PATH}
-	install -T ${REPO}/bin/burrow ${BIN_PATH}/burrow
+	install ${REPO}/bin/burrow ${BIN_PATH}/burrow
 
 # build burrow with checks for race conditions
 .PHONY: build_race_db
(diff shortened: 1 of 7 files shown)


compleatang:
Nice. Thanks!

seanyoung:
@silasdavis @compleatang 

This was merged into develop. The whole reason we got rid of develop is to prevent these sorts of issues.

compleatang:
my fault. fixed here ->  #1455

bzp99:
> This was merged into develop. The whole reason we got rid of develop is to prevent these sorts of issues.

Should I have created the PR into master? _(Then the contributing guide should be updated)_

silasdavis:
> The whole reason we got rid of develop is to prevent these sorts of issues

The new `develop` was never meant to do the same job as the new `develop`, clearly it is a poor choice of name, but as @bzp99 correctly points out the contributing guidelines are wrong and need to be updated.

I've addressed both things here:

https://github.com/hyperledger/burrow/pull/1457

PR #1480, opened 2021-04-02
Fix MutableForest concurrency

silasdavis:
`CallSim` and `CallCodeSim`.

State is separated into a mutable write side and an immutable read side
(from last commit/height) via RWTree and then via MutableForest in an
MVCC style manner. Reads should hit the last committed version. However
there were two bugs in TransactServer's `CallTxSim` and `CallCodeSim`:

1. During the execution of a simulated call if a block is committed
reads can be made from consecutive blocks and so may not see a
consistent view of the state.

2. `ImmutableForest`'s `loadOrCreateTree` was unsynchronised and
committed a multitude of sins: creating and returning multiple trees for the
same prefix when it should ensure one instance exists in cache and most problematically allowing access to uninitialised trees by pushing into LRU
cache before `Load`ing the tree from DB.

The fix for 1 is to pull a snapshot `ImmutableForest` when Call*Sim are
called which is fixed to the latest height at the time of the call.

The fix for 2 was to synchronise `loadOrCreateTree` and only push
initialised trees into the `treeCache`.

simulated_call_test.go adds a test that triggers the second bug
(requires a cold cache and a prefix with previous existing versions)


acm/acmstate/state_cache.go
```diff
@@ -262,21 +262,26 @@ func (cache *Cache) get(address crypto.Address) (*accountInfo, error) {
...
-			if err != nil {
-				return nil, err
-			}
-			accInfo = &accountInfo{
-				account: account,
-				storage: make(map[binary.Word256][]byte),
-			}
-			cache.accounts[address] = accInfo
-		}
+	if accInfo != nil {
```
silasdavis, line 265: this just flattens the ifs a bit for readability, no bug here

execution/simulated_call_test.go
```diff
@@ -0,0 +1,130 @@
...
+		delegate[7] = byte(i)
+		_, _, err = st.Update(func(up state.Updatable) error {
+			return setDelegate(up, delegate)
+		})
+		require.NoError(t, err)
+	}
+
+	// This is important in order to illicit the former bug - we need a cold LRU tree cache in MutableForest
+	st, err = state.LoadState(db, st.Version())
+
```
silasdavis, line 101: Need error check here

silasdavis:
I'm going to push a few other changes to this, namely:

- Add a concurrency test to forest_test.go so we aren't just testing it indirectly
- Add a bunch of comments to forest.go 
- Rather than have a complicated contract of 'single writer multiple readers' I'm going to make forest completely concurrency safe. Since reads are already mostly lock-free and uncontended write mutexes are low overhead we just might as well and avoid possible footguns. I think there may be some other subtle read-side issues in forest.

PR #1486, opened 2021-05-15
Get rid of excessively long and generally useless EVM trace from call

silasdavis:
errors


silasdavis:
:smiley_cat:

PR #1488, opened 2021-05-19
Fixed RLP encoding extra leading zeros on uint64

maoueh:
When a uint64 fits perfectly in a multiple 8 bits, a leading zero was being added to the encoded value, for example 233 was encoded (prefix elided) as `00 e9` while it should have been `e9`

Simply changed the way the amount of byte required is computed so the right amount of leading zeros is correctly removed after encoding.

For uint64 value <= 255, this will even save an allocation since we will more often enter the `encodeUint8` case which is faster (theorically).


silasdavis:
Delightful to see bug fixes like this come in, thanks very much :)

PR #1489, opened 2021-05-20
Fix finitely bounded event streams in the presence of empty blocks.

silasdavis:
ExecutionEventsServer assumes every block will be represented in
state, but this is not true for empty blocks.


deanvanniekerk (approved):

silasdavis:
:smiley_cat: pending getting @DeanVanNiekerk in Hyperledger GitHub org

PR #1505, opened 2021-07-23
Make `deploy` take an object argument with options and library addresses

silasdavis:
for easier generic deployment code and reflection by downstream users


silasdavis:
:smiley_cat:

PR #1525, opened 2021-12-04
fix encoding error of dynamic objet after uint[]

chenfengjin:
fix call encoding  bug 

call contract method increase of
`

    pragma solidity >=0.0.0;

    contract Counter {
        constructor() public{
        }
       function increase(uint256[] memory _tokenIds,string memory _uri) public {

        }
}
`

with parameters
 `["201","202",'203"],"hello"`
 will cause execution reverted.

it is encoded as 

`0b88819400000000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000000060000000000000000000000000000000000000000000000000000000000000000300000000000000000000000000000000000000000000000000000000000000c900000000000000000000000000000000000000000000000000000000000000ca00000000000000000000000000000000000000000000000000000000000000cb000000000000000000000000000000000000000000000000000000000000000568656c6c6f000000000000000000000000000000000000000000000000000000`

which is different from ehtereum encoding result 
`0x0b888194000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000c0000000000000000000000000000000000000000000000000000000000000000300000000000000000000000000000000000000000000000000000000000000c900000000000000000000000000000000000000000000000000000000000000ca00000000000000000000000000000000000000000000000000000000000000cb000000000000000000000000000000000000000000000000000000000000000568656c6c6f000000000000000000000000000000000000000000000000000000`

### DIFF
--- a/execution/evm/abi/abi_test.go
+++ b/execution/evm/abi/abi_test.go
@@ -139,6 +139,13 @@ func TestPacker(t *testing.T) {
 			"arrayOfBoolsPack",
 			append(pad([]byte{1}, 32, true), pad([]byte{0}, 32, true)...),
 		},
+		// test dynamic objet  uint[]
+		{
+			`[{"inputs":[],"stateMutability":"nonpayable","type":"constructor"},{"inputs":[{"internalType":"uint256[]","name":"_tokenIds","type":"uint256[]"},{"internalType":"string","name":"_uri","type":"string"}],"name":"increase","outputs":[],"stateMutability":"nonpayable","type":"function"}]`,
+			[]interface{}{[]string{"201", "202", "203"}, "hello"},
+			"increase",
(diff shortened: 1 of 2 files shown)


seanyoung:
The change looks right. It would be nice to have a test case for this though.

silasdavis:
Thanks @chenfengjin ! May I ask, how did you come across the issue, and to be even nosier what is your use case for Burrow?

chenfengjin:
we developed a blockchain system named XuperChain and we use burrow  as EVM execution engine for it. you can find more information  about this system [here](https://github.com/xuperchain/xuperchain).
Last week one of our user reported this problem to us.

chenfengjin:
unfortunately, I found out that [truffle test](https://github.com/hyperledger/burrow/runs/4456567685?check_suite_focus=true) always fail, but it is not because of my change as I found out that it fail even with nothing change [in my fork](https://github.com/chenfengjin/burrow)

